PR #4, opened 2018-08-27
Update kalman filter

arnavs:
The Plots output has changed a bit, so that some of the labelings in the old plots are lost. Also, the exercises can't be updated until `QuantEcon` is ready.


arnavs:
Sorry, thought this was my branch.

PR #10, opened 2018-08-27
WIP: Style guide

jlperla:


jlperla:
@Nosferican @arnavs @sglyon @cc7768 @jstac 
Here is a starting point for the rough notes of a styleguide.  If you have any comments, post them here for a discussion.  It is very incomplete, but I wanted to give people a chance to provide thoughts before too much rewriting of the style occurs.

I think this is intended to capture discussions that occured in https://github.com/jstac/QuantEcon.lectures/issues/1176 and https://github.com/jstac/QuantEcon.lectures/issues/1023 and https://github.com/jstac/QuantEcon.lectures/issues/1031   though you will see I read them in a rather biased way :-)

One of the things at the top is also the core principles of the style, which I think is important to establish.  These style guidelines would *not* be appropriate for writing a package, for example, but are probably perfectly solid for real research scripts which uses a package.

style.md
```diff
@@ -0,0 +1,281 @@
...
+3. Keep things as **close to the whiteboard math** as possible, including in the code structure and notation
+4. Ensure that all **code can be copied and pasted without modification** into functions for performance and modification without changes to scoping (e.g. no `local` or `global` ever required)
+5. **Avoid type annotations** unless they are required for dispatching
+6. **Avoid creating custom types** unless it is absolutely necessary
+7. Beyond the [Julia Style Guide](https://docs.julialang.org/en/v1/manual/style-guide/), avoid unnecessary whitespace lines and redundant comments
+8. Don't use fancy features and control flow from Julia - unless it makes the code look closer to the math
+9. Avoid both micro-optimizations and coding patterns that pessimize (i.e. poor performance with no benefit in code clarity)
+
+We want users to be able to say _"the code is clearer than Matlab, and even closer to the math"_.  If they want to write their own packages, they will need to learn much more about Julia, but most never will need to.
```
Nosferican: I would delete the last part.

style.md
```diff
@@ -0,0 +1,281 @@
...
+variable2 = 2.0
+
+# ACCEPTABLE BUT OFTEN UNNECESSARY
+A = [1 2;
+     3 4]
+```
+- **Do not** use docstrings in any lecture code - except when explaining packages and the `?` environment.
+- **Feel free** to use the `⋅` unicode symbol, i.e. `\cdot<TAB>` instead of `dot( , )`
+- **Use `LaTeXStrings.jl`** for all latex literals, i.e. `L"\hat{\alpha}"` instead of `""\$\\hat{\\alpha}\$""`
+- **Prefer** `in` to `∈`
```
Nosferican: Isn't this an exception to the rule?

style.md
arnavs: FWIW, I think it makes sense. It's a good hedge against people who raise objections like "why don't you use `{XYZ Fancy Coding Paradigm}`? The explanation is that we understand how good, performant Julia is written, but are sometimes making a conscious departure.

style.md
```diff
@@ -0,0 +1,281 @@
...
+
+- **Avoid:** the short-circuiting pattern unless it is very clear.  This is especially true for assertions which shouldn't be in the normal control flow
+```julia
+# BAD!
+a < 0 && error("`a` parameter invalid")
+
+# GOOD!
+@assert a < 0 
+```
+- **Do not** use a `;` to call functions with keyword arguments unless it is required.  i.e. prefer `range(1, stop=5)` to `range(1; stop=5)`
```
Nosferican: I am pretty sure that's covered by Julia style guide already.

style.md
```diff
@@ -0,0 +1,281 @@
...
+x
+
+# Best!
+x = [2.0 * i^2 for i in 1:N]
+
+# or
+f(i) = 2.0 * i^2
+x = f.(1:5) #Use broadcasting
+```
+- Prefer `eachindex` to accessing the sizes of matrices.
```
Nosferican: This should probably be of vectors.

style.md
jlperla: The rule of using unicode?  Maybe, but not necessarily.  The difference is that I am suggesting using unicode for things that match the whiteboard math in the paper.  If you had a set `A` and wanted to see if `a` was in the set,  and the math in the document said `a ∈ A` then I think it would make sense to use exactly that in the code.

However, when thinking about for loops and comprehensions, it is code rather than math.  I don't have very strong feelings.  The `∈` just seems excessivly fancy to me (kind of like when people you a ∀  symbol in a paper when "for all" would have read perfectly well.

style.md
```diff
@@ -0,0 +1,281 @@
...
+
+# BETTER!
+for i in eachindex(x)
+    y[i] = x[i] + 2
+end
+y
+
+# GOOD! #No way to preallocate y (although easier ways to write)
+x = rand(10)
+y = similar(x, 0) #Empty of same type as x
```
Nosferican: Probably simpler `y = similar(x)`

style.md
```diff
@@ -0,0 +1,281 @@
...
+end
+n
+
+# BEST!
+sum(xval -> xval^2, x) #i.e. transform each x and then reduce
+```
+
+- **Avoid `range` when possible** and use the `1.0:0.1:1.0` style notation, etc.  Getting rid of the `linspace` is a pain, but use ranges if the exact top doesn't matter.
+```julia
+Base.OneTo(5) # BAD!
```
Nosferican: That was just me... probably not worth mentioning it. Might generate more confusion.

style.md
jlperla: That wouldn't be the same, since it would allocate a vector of `10`.

This leads to a big style quetsion I have.  Even though you don't always need to give the size of the vector when using `similar`, is it helpful to try to?

style.md
```diff
@@ -0,0 +1,281 @@
...
+7. Beyond the [Julia Style Guide](https://docs.julialang.org/en/v1/manual/style-guide/), avoid unnecessary whitespace lines and redundant comments
+8. Don't use fancy features and control flow from Julia - unless it makes the code look closer to the math
+9. Avoid both micro-optimizations and coding patterns that pessimize (i.e. poor performance with no benefit in code clarity)
+
+We want users to be able to say _"the code is clearer than Matlab, and even closer to the math"_.
+
+## Naming Conventions, Comments, etc.
+
+- **Use unicode** where possible in names so that symbols match the math in the document
+- **Be careful** about unicode glyphs and symbols which may not be available in the default REPL, Jupyter, etc. for all platforms.  **TODO**
```
arnavs: I think we should reword this a bit, because we're saying on the one hand to use this wherever possible (including, as written, for things like `tilde` and `bar` and `hat`), and on the other to be careful.

style.md
```diff
@@ -0,0 +1,281 @@
...
+
+- **Use external packages** whenever possible, and never rewrite code that is available in a well-maintained external package (even if it is imperfect)
+- The following packages can be used as a dependency without any concerns: `QuantEcon, Parameters, Optim, Roots, Expectations, NLsolve, DataFrames, LaTeXStrings, Plots, ....`
+- **Do** use `using` where possible (i.e. not `import`), and include the whole package as opposed to selecting only particular functions or types.
+- **Prefer** to keep packages used throughout the lecture at the top of the first block (e.g. `using LinearAlgebra, Parameters, QuantEcon`)  but packages used only in a single place should have the `using` local to that use.
+    - If `Plots` is only used lower down in the lecture, then try to have it local to that section to ensure faster loading time.
+- **Always seed random numbers** in order for automated testing to function.
+
+## Work in Progress Discussions
+1. When and how do we use `eachindex` with multidimensional arrays you want to iterate through?
```
Nosferican:
```
for i in 1:size(X, 1)
    for j in 1:size(X, 2)
    end
end
```
Always forget which one goes first, but it is in the docs.

style.md
```diff
@@ -0,0 +1,281 @@
...
+2. How best to stack arrays and unpack them for use with solvers/etc.?
```
Nosferican: That's very MATLAB-ish. I think most of Optim-land doesn't require it anymore.

style.md
jlperla: Is there a way to do it cleanly with `eachindex` or something similar, or is that the preferred way?

style.md
```diff
@@ -0,0 +1,281 @@
...
+- **Prefer** to keep packages used throughout the lecture at the top of the first block (e.g. `using LinearAlgebra, Parameters, QuantEcon`)  but packages used only in a single place should have the `using` local to that use.
+    - If `Plots` is only used lower down in the lecture, then try to have it local to that section to ensure faster loading time.
+- **Always seed random numbers** in order for automated testing to function.
+
+## Work in Progress Discussions
+1. When and how do we use `eachindex` with multidimensional arrays you want to iterate through?
+2. How best to stack arrays and unpack them for use with solvers/etc.?  
+3. Simple error handling of reasonable failures (e.g. returning a union with `Nothing`, etc.)  I am not sure we want to teach them about exceptions, for example.
+4. Can we really use Latex for labels?  Is it stable enough with various backends?  If so, I think that LatexStrings is the way to go.
+5. Lets say `x` is vector of length `N`.  Is `y = similar(x, N)` or `y = similar(x)` a better way to denote the preallocation of the same size?  I kind of like the first in this case, but I think we can use our judgement.
```
Nosferican: `y = similar(x)` the `N` is just redundant.

style.md
sglyon: @Nosferican rule of thumb is to always make the early index the inner most loop.

So with `x = rand(2, 2, 2)` it would be most efficient to iterate `for k in 1:size(x, 3), j in 1:size(x, 2), i in 1:size(x, 1)`

style.md
jlperla: Alas, I am not sure that is true outside of `JuMP`.    I am running into it every day when doing the DAE problems, solving bellman equations with different discrete states, etc.

You will also see it if you collapse some of the dynamic programming problems here, such as the McCall model.

style.md
Nosferican: `CartesianIndices`, but probably not something taught at beginner's level. If possible a helper and `mapreduce` or a nested `for` loop.

style.md
arnavs: A few other things which come to mind: 

1. A line about using `Missing` from Julia Base, instead of `"N/A"`, `NaN`, `Inf`, or something like that. 

2. A line saying don't use the ternary operator (or avoid it if you can), since I know you don't like it. 

3. A line about meaningful names for variables and functions.

style.md
Nosferican: I would rather keep it (pro  ∈). At least in my math classes no one would ever write "for all" and use ∀, or such as ∋ or subject to `:` or given `|`. It reads for each element in that collection or for each member of... From a coding/reading it saves a character length in the line.

How about `≠`, `∉`?

style.md
jlperla: > So with x = rand(2, 2, 2) it would be most efficient to iterate for k in 1:size(x, 3), j in 1:size(x, 2), i in 1:size(x, 1) But as hinted here, eachindex will take care of that for us!

OK, I think I have added something there now.  But @sglyon  How exactly do you do it with `eachindex` in two dimensions, and still get access to the indices themselves?  If there is no clean way to do it, we can just use the julia style guide fallback and say nothing.

Nosferican:
```
X = rand(1,2,2);
for idx ∈ CartesianIndices(X)
       println(idx)
       end
CartesianIndex(1, 1, 1)
CartesianIndex(1, 2, 1)
CartesianIndex(1, 1, 2)
CartesianIndex(1, 2, 2)
```

style.md
jlperla: My issue is that in a construction such as
```julia
for i ∈ 1:10
  @show(i)
end
```
it is inherently code, not the for loop isn't really math.  If it was implementing an actual equation, then I would be for the symbol.  Same with ≠, ∉

The other issue is that students may get confused on how to type it, especially those ones.

But let me ask a better quetsion:  What do the main julia docs do?  If they use ∈ instead of `in` and `≠` instead of `!=`  ... in code blocks, not in exact mappings to algebra of an equations... then I am all for going fancy.

Nosferican: No stance. They are aliases so they are the same operator. Biggest difference is probably,
```
julia> in.(1, 1:3)
3-element BitArray{1}:
  true
 false
 false

julia> .∈(1, 1:3)
3-element BitArray{1}:
  true
 false
 false
```

style.md
jlperla: The problem here is that we want to be able to get people to write code that works with generic code and types.

Using `zeros(x)` looks nice and clean, but it is really just doings `zeros{Float64}(x)`.

After we get rid of all of the types from the function definitions, the only place that they will ever have an issue with types seems to be in preallocations.  If they use `y = similar(x)` instead of `y = zeros(length(x))`  then it will work for `x` an Dual number, for GPU thigns, arbitrary precision numbers, interval arithmetic, etc.

style.md
jlperla: Once we test for the fonts we can.

style.md
Nosferican: `similar` copies the type which is probably what you want, otherwise it would just use `zeros(T, N)`

Nosferican: The `Missing` and `Nothing` is covered on the Julia style guide.

jlperla: Exactly.  No need to repeat those sorts of things.  The only reason to add something on that for this code is if we find an error handling pattern we think is simple and teachable, but lets wait on that.

jlperla: Yup, which is why I want to train people to use `similar(x)`  `similar(x, N, N)` etc. whenever possible.  It skips forcing in `Float64`  when you don't want one.

Nosferican:
Something I saw in the code was:
```
struct Name
    a :: T1
    b :: T2
end
```
I prefer
```
struct Name
    a::T1
    b::T2
end
```
As for args/vals
```
plot(data, labels=L"x")
```
I prefer
```
plot(data, labels = L"x")
```

style.md
arnavs: Gotcha.

sglyon:
I also vote no whitespace around `::`

But I would vote for no space around `=` for keyword args

style.md
jlperla: @sglyon your thoughts? I don't have very strong opinions, but lean towards `in` for non math code

cc7768: I am in favor of `in` rather than `∈` when writing loops or other code based statements but am also in favor to using `∈` when it allows code to look more like whiteboard math.

jlperla: @Nosferican Sorry, I think you are outvoted on this one.  Things that are inherently code and control flow should stay ascii.  So `in` and `!=`  instead of `∈, ≠`.

I added a point about using those unicode symbols for set inclusion.

jlperla:
> But I would vote for no space around = for keyword args

I don't have a strong feeling.  My guess is that for a lot of whitespace we will just be able to say "Run it through a beautifier" within the next 6 months.  e.g. https://discourse.julialang.org/t/julia-code-formatter-code-beautifier/11485/7

jlperla:
Do you guys feel like this is decent enough to merge for now?  If this captures the main things (with minimal repetition from the standard Julia ones) then maybe it is worth it to get started.

jlperla:
Merging.  Post issues or make PRs if you want things changed.  When things like the Julia code formatter are up and running, we may be able to remove some of the guidelines regarding whitespace/etc.

PR #26, opened 2018-08-28
Update Julia by Example

Nosferican:


arnavs:
Can confirm that this is undeprecated. But I want to hold off on merging until we figure out a clean, permanent solution to the scoping issue Jesse mentioned in #39.

Additionally, I think this syntax: 

```
using Pkg: add, installed
"Distributions" ∈ keys(installed()) || add("Distributions")
```

should either be explained or hidden. I think @jlperla's goal was to make these lectures similar to an Apple demo --- showing you no more than you need to get on board.

jlperla:
Yeah. I am generally loathe to use anything relying on short circuiting logic at all. It is pretty advanced for new users, and even confusing for good programmers coming from many languages. 

But I suspect that code is just temporary until we have the Manifest stuff working

PR #28, opened 2018-08-28
Update Julia Arrays

Nosferican:


rst_files/julia_arrays.rst
```diff
@@ -268,41 +212,32 @@ In such cases, we don't care about the distinction between row and column vector
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 We've already seen some functions for creating arrays
 
 .. code-block:: julia
 
-    eye(2)
-
-
+    Matrix{Float64}(I, 2, 2)
```
arnavs: I think using `Matrix{Float64}` as a function would be confusing (especially since it takes 3 arguments). I think `ones(3, 3)` could work here as a different example.

rst_files/julia_arrays.rst
```diff
@@ -268,41 +212,32 @@ In such cases, we don't care about the distinction between row and column vector
...
 
 
-
-
 You can create an empty array using the ``Array()`` constructor
 
 .. code-block:: julia
 
-    x = Array{Float64}(2, 2)
+    x = Array{Float64}(undef, 2, 2)
```
arnavs: This is a direct port of the relevant QuantEcon...but I still think it could be confusing. Will defer to @jlperla.

jlperla: I think we can hold off on most stylistic changes for now.

arnavs: OK, got it.

Nosferican: I kept that one since it is teaching how to use the constructor. The next lines go about how those are garbage values, etc.

rst_files/julia_arrays.rst
Nosferican: Is important to teach the current way of getting an identity matrix.
```
I # best case scenario
diagm(0 => ones(n))
Matrix{T}(I, n, n)
Diagonal(ones(n))
```
The `I` is covered, but we could choose ones of the other ones.

PR #29, opened 2018-08-28
Update Types and Methods

Nosferican:
- Made a small change to fix an issue in the material
- We could add something about constructors eventually

### DIFF
--- a/rst_files/types_methods.rst
+++ b/rst_files/types_methods.rst
@@ -41,8 +41,6 @@ Types and Multiple Dispatch
 ===========================
 
 
-
-
 Common Types
 ---------------
 
@@ -52,7 +50,6 @@ In Julia all objects have a type, which can be queried using the ``typeof()`` fu
 .. code-block:: julia
 
     typeof(0.5)
-
 
 
 .. code-block:: julia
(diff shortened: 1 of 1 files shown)


rst_files/types_methods.rst
```diff
@@ -495,14 +447,13 @@ If the interpreter can't find a match in immediate parents (supertypes) it proce
 
 * If not, we get an error
 
-This is the process that leads to the following error 
+This is the process that leads to the following error
```
arnavs: Doesn't yield the desired error, because this block inherits `+(x::Integer, y::String)` from an earlier code block. Will eventually need to replace with a valid example.

rst_files/types_methods.rst
```diff
@@ -745,7 +672,6 @@ For the distribution ``ϕ`` we'll assign a ``Distribution`` from the `Distributi
     m = AR1(0.9, 1, 1, Beta(5, 5))
 
 
-
 In this call to the constructor we've created an instance of ``AR1`` and bound the name ``m`` to it
 
 We can access the fields of ``m`` using their names and "dotted attribute" notation
```
arnavs: Should probably put something in here eventually about how direct access is a bad idea.

PR #31, opened 2018-08-28
Update Julia Libraries

Nosferican:


jlperla:
> Maybe eventually we could put in a quick note about what a Symbol is.

We are going to completely reorganize this sort of material after we get it ported to v1.0 and rewrite, so lets not worry about content additions yet.

arnavs:
@jlperla OK, thanks for letting me know. I had figured that this stuff would be useful to come back to when we start on the style edits.

PR #32, opened 2018-08-28
Update Linear Algebra

Nosferican:


rst_files/linear_algebra.rst
```diff
@@ -278,35 +264,29 @@ The expression :math:`\| x - y\|` is thought of as the distance between :math:`x
...
 .. code-block:: julia
 
     dot(x, y)               # Inner product of x and y
 
 
-
 .. code-block:: julia
 
-    sum(x .* y)             # Gives the same result
+    sum(i * j for (i, j) ∈ zip(x, y))             # Gives the same result
```
jlperla: This is way too verbose.  I would rather have an unnecessary allocation then something so complicated.

rst_files/linear_algebra.rst
```diff
@@ -278,35 +264,29 @@ The expression :math:`\| x - y\|` is thought of as the distance between :math:`x
...
 
-
 .. code-block:: julia
 
-    sqrt(sum(x.^2))         # Gives the same result
-
-
-
-
+    sqrt(sum(abs2, x))         # Gives the same result
```
jlperla: I think that leaving both of these examples in there is probably a good idea.

Nosferican: I used `sum(_ -> _^2, x)` and other variants throughout the lectures to introduce a new convenient function.

rst_files/linear_algebra.rst
Nosferican:
```
sum(*, zip(x, y)) # errors out
```
but this works
```
sum(prod, zip(x, y))
```

PR #33, opened 2018-08-28
Update Orthogonal Projections

Nosferican:


Nosferican:
They are aware of the issue and is a feature that they haven't included, but are working on it.

PR #34, opened 2018-08-28
Update LLN and CLT

Nosferican:


arnavs:
Generally looks great. A few things: 

1. What's going on with the title of this plot? (Ignore the `dvipng` issues, those are probably missing some command-line tool on my local machine). 

screen shot 2018-08-28 at 11 25 54 am

2. For some reason, our version (below) of [this plot](https://lectures.quantecon.org/_images/clt3d.png) looks a lot worse. Some issues in the plotting library? Maybe we can take another look. 

3. The `KernelDensity` package is throwing some kind of `Warning: Could not import LinearAlgebra into Interpolations.jl` error. Doesn't seem to break anything...but good to keep an eye on.

PR #42, opened 2018-08-29
Undeprecate Asset Pricing I

arnavs:
Doesn't perform any stylistic changes; just a clean port. @Nosferican, once you give the go-ahead, I'll move on to the other two lectures in the mini-series as we discussed.


Nosferican:
Great! Good for the next ones.
---
Note to self, check for `Random.seed!` in lectures since some use it consistency and others do not for when re-writing it.

PR #43, opened 2018-08-29
Undeprecate Asset Pricing II 

arnavs:
As before, strictly functional update. Also eliminated the one-off inclusion of `lucastree.jl`. 

(Edit: I believe the only functional update is going from `linspace -> range`, and modifying the object to accept `grid::AbstractRange`).


rst_files/lucas_model.rst
```diff
@@ -414,12 +414,109 @@ Let's try this when :math:`\ln y_{t+1} = \alpha \ln y_t + \sigma \epsilon_{t+1}`
...
+
+        # == build a grid with mass around stationary distribution == #
+        ssd = σ / sqrt(1 - α^2)
+        grid_min, grid_max = exp(-4 * ssd), exp(4 * ssd)
+        grid = range(grid_min, stop = grid_max, length = grid_size)
+        
+        # == set h(y) = β * int u'(G(y,z)) G(y,z) ϕ(dz) == #
+        h = similar(grid)
+        for (i, y) in enumerate(grid)
+            h[i] = β * mean((y^α .* shocks).^(1 - γ))
```
Nosferican: Shouldn't this have `mean(x, dims = dims)`?

arnavs: Julia didn't complain but let me take a look.

Nosferican: I think is fine, but didn't see a `using Statistics`.

arnavs: This collection is one-dimensional, so it doesn't make a semantic difference:

```
julia> mean([1, 2, 3, 4])
2.5

julia> mean([1, 2, 3, 4]')
2.5

julia> mean([1 2 3 4])
2.5
```

What this should have, though, is `using Statistics`. But we escaped, probably because `Distributions` exports its own version of `mean` that extends the base one.

Nosferican: I guess it should be fine in this context.

PR #46, opened 2018-08-30
Test Orthogonal Projections

arnavs:
Not much code in this one. Just verified that the three different projection schemes come out to the same result.

### DIFF
--- a/rst_files/orth_proj.rst
+++ b/rst_files/orth_proj.rst
@@ -848,6 +848,11 @@ Here's a function that computes the orthonormal vectors using the GS
 algorithm given in the lecture.
 
 .. code-block:: julia
+    :class: Test 
+
+    using Test # Put this before any code in the lecture. 
+
+.. code-block:: julia
 
     using LinearAlgebra
 
@@ -891,6 +896,14 @@ by the columns of :math:`X`.
 
     Py1 = X * inv(X'X) * X' * y
 
+.. code-block:: julia 
+    :class: test 
+
+    @testset "Test Py1" begin
+        @test Py1[1] ≈ -0.5652173913043479 atol = 1e-10
+        @test Py1[2] ≈ 3.2608695652173916 atol = 1e-10
+        @test Py1[3] ≈ -2.217391304347826 atol = 1e-10
+    end 
 
 Now let's orthogonalize first, using Gram--Schmidt:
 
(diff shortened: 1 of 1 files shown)


rst_files/orth_proj.rst
```diff
@@ -919,6 +938,12 @@ basis by QR decomposition and project once more.
 
 .. code-block:: julia
 
-    Py2 ≈ Q * Q' * y
+    Py3 = Q * Q' * y
+
+.. code-block:: julia 
+    :class: test 
+    @testset "Test Py3" begin 
+        @test norm(Py3-Py1) ≈ 0 atol = 1e-10
```
Nosferican: Wouldn't this just be `@test Py1 ≈ Py3 atol = 1e-10`?

arnavs: Maybe they're equivalent, but that was the quantity I used in the REPL.

Nosferican: We probably want to test `a  ≈ b` since there might be lectures that don't use `LinearAlgebra` and we don't want to include that dependency as a `target`.

jlperla:
Do we really need all of these `atol = 1e-10` everywhere?  Isn't the default tolerance enough?  I would prefer to drop them if possible.

One thing, though, is that you can never test that something `@test x ≈ 0`  without an `atol` as it can't figure out the relative tolerance between the two sides.  We will need to discuss that in the notes at some point.

jlperla:
I think we should have `using LinearAlgebra` at the top of every lecture.  Julia without that is pretty confusing for users who start playing with the notebook.

arnavs:
@jlperla The default tolerance for `isapprox` is full machine precision (`1e-16`), I believe. I put these in as a precautionary measure (maybe we're not printing the full float, maybe there are differences with different machines, etc.). But they could be redundant, and we could probably remove them at some point.

Nosferican:
Having `LinearAlgebra` at least at this stage will break some lectures given `I` and a few other namespace conflicts. Might be sensible during the re-write...

Nosferican:
I know for some with QuantEcon given a naive (in the lectures) and a high precision one (in the package) I think I needed `1e-6` or `1e-8`.

jlperla:
> The default tolerance for isapprox is full machine precision (1e-16),

That isn't true.  It uses the machine epsilon, but comparing floating points adjusts for the scale.  See https://docs.julialang.org/en/v1.0.0/base/math/#Base.isapprox

Try to avoid them if possible and just add in extra numbers if necessary, but not a big deal.

jlperla:
> Having LinearAlgebra at least at this stage will break some lectures given I and a few other namespace conflicts. Might be sensible during the re-write...

Sure, it can wait.  But it also is only a problem if you use the `I` as a uniform scaling.  See
```julia
julia> using LinearAlgebra

julia> I = 5
5

julia> I
5
```

arnavs:
OK, I stripped the tolerances from the tests and there were no issues. @Nosferican is this good to merge?

rst_files/orth_proj.rst
arnavs: That's a good point. Will do this as well.

Issue #41, opened 2018-08-28
Global Regression Test Tracker


arnavs:
For a given lecture, in between the undeprecation and the rewriting/refactoring, we want to insert regression tests to ensure that the output does not change. 

# Guidelines

Some thoughts, based on discussions with @jlperla:

- When testing arrays, try not to test the bounds, but rather some arbitrary interior value. 
- In general, try to test values that depend on many other values, since it's unlikely that errors in those components will exactly cancel. Something like a checksum. 
- ~~We should put these tests in their own blocks with a special comment (`# jl-test`, I believe), so that we can later efficiently delete all blocks that contain that comment.~~ See README for this now. 
- ~~For plots, the testing procedure is still TBD. Perhaps we can eyeball it, or maybe serialize the plot object and save it. Not sure what a good approach is here.~~ See Jesse's comment below for how to handle this. 

# Tracker

## Programming in Julia 
- [ ] [Setting up your Julia Environment](https://lectures.quantecon.org/jl/getting_started.html)
- [ ] [An Introductory Example](https://lectures.quantecon.org/jl/julia_by_example.html)
- [ ] [Julia Essentials](https://lectures.quantecon.org/jl/julia_essentials.html)
- [ ] [Vectors, Arrays, and Matrices](https://lectures.quantecon.org/jl/julia_arrays.html)
- [ ] [Types, Methods, and Dispatch](https://lectures.quantecon.org/jl/types_methods.html)
- [ ] [The Need for Speed](https://lectures.quantecon.org/jl/need_for_speed.html)
- [ ] [Plotting in Julia](https://lectures.quantecon.org/jl/julia_plots.html)
- [ ] [Useful Libraries](https://lectures.quantecon.org/jl/julia_libraries.html)

## Tools and Techniques 
- [x] [Linear Algebra](https://lectures.quantecon.org/jl/linear_algebra.html)
- [x] [Orthogonal Projections and their Applications](https://lectures.quantecon.org/jl/orth_proj.html) (#46)
- [x] [LLN and CLT](https://lectures.quantecon.org/jl/lln_clt.html) (#47)
- [x] [Linear State-Space Models](https://lectures.quantecon.org/jl/linear_models.html) (#48)
- [x] [Finite Markov Chains](https://lectures.quantecon.org/jl/finite_markov.html) (#49)
- [x] [Continuous State Markov Chains](https://lectures.quantecon.org/jl/stationary_densities.html) (#50)
[...]

PR #48, opened 2018-08-30
Test Linear Models

arnavs:
Add tests to the relevant lecture. Partial resolution to issue #41.

### DIFF
--- a/rst_files/linear_models.rst
+++ b/rst_files/linear_models.rst
@@ -1377,10 +1377,10 @@ the bars are at dates 10, 50 and 75
 Solutions
 ==========
 
-.. code-block:: julia 
-    :class: test 
+.. code-block:: julia
+    :class: test
 
-    using Test 
+    using Test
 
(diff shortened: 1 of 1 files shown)


arnavs:
Note: There's one line of this lecture which we should flag for re-writing. In Exercise 3, we define `I = 20`, and then modify some matrix with `foo ./ I`. I think that's error-prone (or at least bad practice) now that `LinearAlgebra.I` refers to the identity matrix.

jlperla:
`LinearAlgebra.I` is for UniformScaling, not the identity matrix :-)  It actually does matter (for example, in that issue you were having when substracting it with `.-= I`.

Regardless... It is more important to match the math in the lecture notes than anything else right now.  So if `I` is used in the math, and we don't use uniform scaling, then keep it as  the symbol  for the sizes.   We can adjust math down the road to change notation away from the `I` (at which point it makes sense to get rid of that).

PR #54, opened 2018-08-31
Eliminate Literal Includes

arnavs:
Performs the brunt of the housekeeping we discussed on slack; i.e., getting rid of `literalinclude`s for files that are used in only one or two lectures. This should make it easier to refactor the lectures, and hopefully reduce the footprint of the `git` repository. 

@Nosferican, I think a good game plan will be to merge the outstanding PRs, then merge this, and then rebase and continue. ~~I should be done with this by the end of the weekend latest.~~ 

Edit: This is ready to merge now.


jlperla:
What am I supposed to look at?

PR #57, opened 2018-09-05
Test Wald-Friedman

arnavs:
What it says on the tin. An interesting thing to note is that this seems to catch a Jupinx issue. Compare the RST (say, around line 533) to the lecture online. The issue seems to happen when we interspersed text in between a code block and printed side-effects from a function call. Seems like the output doesn't map cleanly, or at all.

Partial resolution to issue #41.

cc: @mmcky for Jupinx.

### DIFF
--- a/rst_files/wald_friedman.rst
+++ b/rst_files/wald_friedman.rst
@@ -171,7 +171,10 @@ The next figure shows two discretized beta distributions in the top panel
 
 The bottom panel presents mixtures of these distributions, with various mixing probabilities :math:`p_k`
 
+.. code-block:: julia 
+    :class: test 
 
+    using Test 
 
 .. code-block:: julia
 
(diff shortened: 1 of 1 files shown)


jlperla:
Did you post the jupinx issue?

arnavs:
@jlperla Not yet, wanted to confirm it here first.

mmcky:
@arnavs it would be helpful if you could add a screenshot to this post so I can see the issue you're referring to. If the issue isn't introduced by the `:class: test` can you please post a screenshot from the comparison site for `wald_friedman.html`

http://compare-lectures.quantecon.org/

rst_files/wald_friedman.rst
```diff
@@ -510,12 +521,16 @@ Here's the code
...
+.. code-block:: julia 
+    :class: test 
 
-Running it produces the following output on our machine
-
-
-
+    @testset "Second Block Tests" begin 
+        @test J1[19] == 0.36
+        @test f1[40] == 0.002163769345396983
```
Nosferican: Use `≈` for comparison with `AbstractFloat`

arnavs: Thanks.

arnavs:
@mmcky Don't think it was. Here's a screenshot. 

screen shot 2018-09-04 at 10 15 30 pm

Nosferican:
Ready to be merged?

arnavs:
I think so.

PR #58, opened 2018-09-05
WIP: Update mccall

Nosferican:
Would be nice to find how to give the contour labels.

### DIFF
--- a/rst_files/mccall_model.rst
+++ b/rst_files/mccall_model.rst
@@ -457,7 +457,6 @@ In particular, let's look at what happens when we change :math:`\beta` and
 .. code:: julia
 
     contourf(c_vals, β_vals, R',
-             fill = true,
              title = "Reservation Wage",
              xlabel = "c",
              ylabel = L"\beta")
--- a/rst_files/mccall_model.rst
+++ b/rst_files/mccall_model.rst
@@ -26,7 +26,7 @@ Overview
 
 The McCall search model :cite:`McCall1970` helped transform economists' way of thinking about labor markets
 
-To clarify vague notions such as "involuntary" unemployment, McCall modeled the decision problem of unemployed agents directly, in terms of factors such as 
+To clarify vague notions such as "involuntary" unemployment, McCall modeled the decision problem of unemployed agents directly, in terms of factors such as
 
 *  current and likely future wages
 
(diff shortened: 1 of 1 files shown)


rst_files/mccall_model.rst
```diff
@@ -354,48 +345,45 @@ Our initial guess :math:`v` is the value of accepting at every given wage
...
-                                      max_iter::Int64=500,
-                                      tol::Float64=1e-6)
+    function compute_reservation_wage(c, β;
+                                      max_iter = 500,
+                                      tol = 1e-6)
 
     # == First compute the value function == #
 
-    v = w_vals ./ (1 - β)
+    v = collect(w_vals ./ (1 - β))
```
arnavs: Is there a reason we need to `collect` here?

rst_files/mccall_model.rst
```diff
@@ -407,10 +395,10 @@ between successive iterates is below `tol`
             v_next[j] = max(stop_val, cont_val)
         end
 
-        error = maximum(abs.(v_next - v))
+        error = maximum(abs(a - b) for (a, b) in zip(v_next, v))
```
arnavs: I'm fine with this. But maybe this is the kind of terse, high-density syntax that @jlperla wants to avoid.

rst_files/mccall_model.rst
Nosferican: It uses an in-place operation afterwards and thus it needs to be a `Vector` and not a `UnitRange`.

rst_files/mccall_model.rst
Nosferican: That whole lectures needs to be re-written, what I might end up favoring is defining a helper function to make it simpler for most lectures... I will think about it.

jlperla: Yeah.  I think the key is that since so much will be rewritten (e.g. this particular example should be replaced with the fixedpoint library!)   it isn't worth doing too much refactoring right now (in either direction... changing the existing code beyond getting it running, or worrying about changes needed to get it running)

But. that said, I think that `error = maximum(abs(a - b) for (a, b) in zip(v_next, v))` is pretty tough and far from the math.  THis is the sort of thing where `error = norm(v_next - v)` is probably appropriate.  Even if it is slightly slower in general, it is mathematically clear.

rst_files/mccall_model.rst
arnavs: Makes sense.

rst_files/mccall_model.rst
arnavs: Yeah, no need to make actual code changes. Just thought I'd bring it up.

PR #59, opened 2018-09-05
Update McMcall II

Nosferican:
Having issues with the LaTeX other than that good.


arnavs:
@jlperla Maybe working with LaTeX and Jupinx is something Adam could help out with.

jlperla:
The latex issue is in the notebook itself, and not with labels?

PR #60, opened 2018-09-05
Update Wald Friedman

Nosferican:
Needs a re-write. Duck-typed.


jlperla:
I think the duck typing with structs is the way to go in the first pass.  Since we will end up using named parameters (with duck typing) in the rewrite, it is the easiest way to get there.

arnavs:
Looks good to me. Mostly the usual fixes with `linspace`, `undef`, and broadcasting.

PR #72, opened 2018-09-07
Test and Undeprecate "Optimal Savings II: LQ Techniques"

arnavs:
~WIP.~

Also: There's a function the v0.6 function relies on (`start(moment_generator)`, where `moment_generator = moment_sequence(lss)`) which isn't in the latest tagged release of QuantEcon (`v0.18.1`). @Nosferican, how should I proceed?

### DIFF
--- a/rst_files/perm_income_cons.rst
+++ b/rst_files/perm_income_cons.rst
@@ -31,26 +31,26 @@ For example, we saw how the model asserts that for any covariance stationary pro
 
 Other applications use the same LQ framework
 
-For example, a model isomorphic to the LQ permanent income model has been used by Robert Barro :cite:`Barro1979` to interpret intertemporal comovements of a government's tax collections, its  expenditures net of debt service, and its public debt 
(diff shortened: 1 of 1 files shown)


Nosferican:
Don't worry about it. If there are non trivial updates I can take care of those and you can review those. In case you feel like it might be quite easy, is a matter of changing it to a for loop and iterate it directly. I updated a case like that a few PR ago. One of the structs has a never finish iterator so you can use `enumerate` and `break` to get the same number of iterations. No need to do state or anything (it simplifies the code too).

arnavs:
Perfect, thanks. Didn't dive too deep onto the internals on this, but that sounds good.

arnavs:
@Nosferican This is ready to review now.

Nosferican:
I didn't see any tests on this lecture, no?

arnavs:
Yeah, the name on this PR should be amended. I fixed it.

PR #81, opened 2018-09-11
Update and Tests Markov Perf

Nosferican:
Request for PyPlot to Plots


arnavs:
@Nosferican made some fixes (`LQ => QuantEcon.LQ`, `1e-8` tolerance => `1e-7`, broadcasting). We can talk about where this is when you get back.

PR #83, opened 2018-09-11
Test and Undeprecate "The Aiyagari Model"

arnavs:


Nosferican:
@arnavs, check comments on my last commit.

arnavs:
@Nosferican for sure, you're correct. I just didn't see the need to optimize the struct fieldtypes when we're going to rip them out anyway.

PR #84, opened 2018-09-13
optgrowth.rst: PyPlot.jl converted to Plots.jl

ajozefiak:


arnavs:
Thanks @ajozefiak, this looks great. Few comments: 

1. Going forward, it would be great if you could target this backend specifically (https://github.com/JuliaPlots/PlotsGR.jl), since that's what we'll be using. 

2. No need to try and match the `LaTeX` strings in the labels, titles, etc., since we'll be ripping those out. 

3. There's some weird issue where our true and approximate value functions seem to be switched. I played around and I don't think this is a result of what you did (i.e., our approximate value function is genuinely below our true one). So I'll merge, but will post an issue. 

screen shot 2018-09-13 at 10 17 33 am

arnavs:
Actually, scratch that re GR. It is the backend we're using, but since it's baked into Plots.jl, there's no reason to do anything special for it. 

And cc @Nosferican for the thing above. I'll confirm that the math is the same, in which case maybe it's due to something like shock stochasticity.

ajozefiak:
Do you want me to completely disregard LaTeX strings? I can include them, the only issue was variable interpolation going from PyPlot to Plots.

PR #85, opened 2018-09-13
Update and Test Default Risk and Income Fluctuations

Nosferican:


arnavs:
We could also probably stand to use more test coverage on this one. Will add to @yolsever's issue.

PR #86, opened 2018-09-13
Update Globalization and Cycles

Nosferican:
Needs Tests and Plots in separate PR.


arnavs:
Small bugfix (`using PyCall`), but otherwise looks great.

PR #94, opened 2018-09-14
Update LQRamsey

Nosferican:
Needs Test and Plots in separate PR.


arnavs:
Looks good to me. Added random seeds for when we get around to testing.

PR #95, opened 2018-09-14
egm_policy_iter: PyPlot to Plots

ajozefiak:


arnavs:
Looks great as well. You left a few LaTeX labels in here, so I went and took them out. Not a huge deal (we'll be doing multiple passes), but if you could rip them out where you find them, that would be great.

PR #97, opened 2018-09-16
smoothing.rst PyPlot to Plots

ajozefiak:


arnavs:
Looks great. One small issue is that we don't want to show labels `y3` and `y4` corresponding to the horizontal axis. Easy fix (which I implemented here) is just to give those lines a null label `label=""`.

PR #99, opened 2018-09-18
Kaan

yolsever:


arnavs:
Hey @yolsever, thanks for this. Looks like the tests are passing, and this adds valuable coverage. I'm making a few small fixes here and there and then this should be good to go. 

Few small things: 

- Make sure you're using proper `rST` syntax. It can be picky about (for example) requiring a line break between the `:class: test` and the code, or not permitting a line break between the `.. code-block:: Julia` and the `:class: test`. 

- Let's try to do one PR per lecture/file. No hard reason for this, but that's the workflow @Nosferican and I have been adopting, and I think it works well. (And then, if you name the branch with the name of the lecture, it's really easy to see what you've worked on).

Other than that looks great; will review tn or tomorrow.

PR #103, opened 2018-09-22
lake_model.rst PyPlot converted  to Plots

ajozefiak:


arnavs:
Perfect, no issues here.

PR #104, opened 2018-09-23
perm_income_cons.rst PyPlot to Plots converted

ajozefiak:


arnavs:
Looks great. We'll maybe do some cosmetic tweaks to the Plots while iterating, but this PR does what's needed.

PR #120, opened 2018-09-25
lqramsey.rst PyPlot to Plots converted, with latex labels stripped to…

ajozefiak:
I believe that gr doesn't support placing legends outside of the subplot as the original PyPlot version had its legends formatted. I have read into one way of working around this, but it requires specifying a layout and making the legend a separate subplot. This could make the lectures look cumbersome so I avoided this.


jlperla:
@ajozefiak  Great.  Less code is better than perfect replication

arnavs:
Plots look great to me. Thanks @ajozefiak, merging now.

PR #126, opened 2018-09-26
Arma plots

ajozefiak:
After completing this one I've realized that the last segment of plotting could be somewhat simplified. Potentially this simplification could be performed later?

### DIFF
--- a/rst_files/arma.rst
+++ b/rst_files/arma.rst
@@ -228,7 +228,7 @@ The next figure plots an example of this function for :math:`\phi = 0.8` and :ma
         times = 0:16
         acov = [ϕ.^k ./ (1 - ϕ.^2) for k in times]
         label = "autocovariance, phi = phi_var"
-        plot!(plots[i], times, acov, color=:blue, lw=2, marker=:circle, markersize=2,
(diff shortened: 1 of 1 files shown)


jlperla:
I guess it depends on how much they would be simplified.  If you think that it would make the code much more clear, then it may be worth doing sooner rather than later.  My main worry would be forgetting about it.

Can you create a new issue (or add to the existing metaissue) with a checklist of the ones you want to review, and a note on the change you would intend to make?

arnavs:
@ajozefiak Just want to confirm that you're not proposing structural changes to code (for example, a lot of our plots are wrapped inside kind of opaque plotting functions like `gen_plots`, and maybe it would be clearer to take them out). Because I think that kind of stuff we can hold off on until we start refactoring in earnest. 

But, like @jlperla said, I don't see any reason not to make any other quick fixes.

ajozefiak:
@arnavs The last snippet of code has functions like plot_simulation and these seem a bit excessive for this one lecture. I can hold off on any such structural changes, but this was one particular lecture that stood out to me.

jlperla:
For sure. I think that kind of excessive complexity will be fixed in the rewriting pass

arnavs:
Plots look great, merging now.

PR #127, opened 2018-09-27
multiplicative_functionals.rst PyPlots to Plots converted (except for a plot title).

ajozefiak:
The last plot that has 6 subplots does not have a title as plot_title is not yet supported in Plots. There are possible workarounds by constructing an additional subplot above the original 6 subplots and making that plot solely a subplot title, though this would make the code substantially clunkier.


arnavs:
No worries about the plot title. I think the moral of the story here is to reproduce as faithfully as we can without compromising readability. If it's unclear what the plot is supposed to be, we can probably just put a header above it in Markdown or something.

PR #144, opened 2018-10-01
arellano.rst has Gadfly to Plots converted

ajozefiak:


arnavs:
Doesn't break the plot, but we're getting a warning `GR: heatmap only supported with equally spaced data.`

PR #151, opened 2018-10-04
Structs => NamedTuples: "A Problem that Stumped Milton Friedman"

arnavs:
~~(WIP: See #152)~~ Opening this PR as an example for @XiaojunGuan. In particular: 

1. Note that we can still use the constructors, even though we've ripped out the types. That way, we don't need to rewrite the intermediate value calculations. 

2. We need to get rid of type annotations on methods, like `current_distribution(wf, p)` (modulo dispatch, as we discussed). The actual method bodies generally don't need to change, since we can use dot notation (`x.y`) to get values from named tuples. 

3. Corollary to (2) is keeping field names consistent. Sometimes this involves nesting named tuples, like we do with the field `sol` of the return of `function WaldFriedman(c, L0, L1, f0, f1; m = 25)`.

4. Sometimes we'll have dummy types, like `struct F0 <: HiddenDistribution end`, that we use for dispatch. I think we can leave these in for now, but we'll rip them out eventually since they abuse the dispatch system (@jlperla?). 

5. We need to change any logic that relies on mutating the state of a `struct`, since that's not possible anymore. See what we did to the function `solve_model!` and `simulate`. 

6. Lastly, we want to make sure tests pass. One way to do this is to run the notebook from the command line (`./run-notebook.sh wald_friedman.ipynb`) and then inspecting the result in `notebooks/`. Or, you could open the notebook in `_build/jupyter` and run all cells. 

Also- if you look at the Julia code vs Python on this and at the plot titles/lecture text, it looks like somewhere along the line the 500 iterations got changed to 5000. This is probably why the lecture takes roughly 10X longer to run on Julia per the build server. So I set it back.

### DIFF
--- a/rst_files/wald_friedman.rst
+++ b/rst_files/wald_friedman.rst
@@ -540,26 +540,6 @@ We shall construct two types that
 
     =#
 
-    mutable struct WFSolution{TAV <: AbstractVector, TR<:Real}
-        J::TAV
-        lb::TR
-        ub::TR
-    end
-
-    struct WaldFriedman{TR <: Real,
-                        TI <: Integer,
-                        TAV1 <: AbstractVector,
-                        TAV2 <: AbstractVector}
-        c::TR
-        L0::TR
-        L1::TR
-        f0::TAV1
-        f1::TAV1
-        m::TI
-        pgrid::TAV2
-        sol::WFSolution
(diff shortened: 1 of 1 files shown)


arnavs:
Closing because of #156

PR #156, opened 2018-10-05
Rewrite of WF lecture

Nosferican:
First major re-write. Feedback is welcomed.


jstac:
@Nosferican I think this looks like a big improvement. I defer to others in terms of knowledge of correct Julia style but to me this looks great.  It's clear, readable, concise and neat.

Thanks!

rst_files/wald_friedman.rst
```diff
@@ -185,57 +168,24 @@ Activate the project environment, ensuring that ``Project.toml`` and ``Manifest.
 
 .. code-block:: julia
 
-  using Distributions, Plots, LaTeXStrings, LinearAlgebra, QuantEcon, Printf, Interpolations
```
sglyon: Did the example source here get deleted, or is the small code block below sufficient to replicate example?

rst_files/wald_friedman.rst
```diff
@@ -185,57 +168,24 @@ Activate the project environment, ensuring that ``Project.toml`` and ``Manifest.
...
-    :class: test
-
-    @testset "First Plot Tests" begin
-        @test mix[7] ≈ 0.005059526866095018
-        @test all(isequal(0.02), f0)
-        @test f1[10] ≈ 0.0011405494355950728
-    end
 
+    plot(plot(reshape(Beta.(1:2:3, 1:2:3), 2, 1), labels = ["f_0", "f_1"], title = "Original Distributions"),
+         plot(reshape(MixtureModel.(Beta, Ref((p -> (p, p)).(1:2:3)), (p -> [p, one(p) - p]).(0.25:0.25:0.75)), 3, 1),
```
sglyon: This code is clever!... maybe a bit too cleverl

Does the syntax of broadcasting over an anonymous function (e.g. `(p -> [p, one(p) - p]).(0.25:0.25:0.75))`) have the potential to scare some of our target audience?

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-
-* First, we use the largest of all the grid points smaller than :math:`p`, and call it :math:`p_i`
-
-* Second, we use the grid point immediately after :math:`p`, named :math:`p_{i+1}`, to approximate the function value as
+Let's code this problem up and solve it.
 
-.. math::
-
-    J(p) = J(p_i) + (p - p_i) \frac{J(p_{i+1}) - J(p_i)}{p_{i+1} - p_{i}}
+We implement the cost functions for each choice considered in the
```
sglyon: This is a break from writing the style adopted in most of the lectures.

In most places on lectures.quantecon.org we have one sentence paragraphs without punctuation at the end. Here we have replaced that with more conventional paragraphs that have multiple sentences separated by punctuation marks.

I'm not casting a vote here for which I think we should have going forward, my purpose is only to point out that the change occurred

This applies to a few places in the diff, but I'll only mention it once

jstac: Right.  Let's stick to one sentence lines with a line break after and no full stop.

That's something we can change later but let's at least be uniform.

jstac:
@mmcky Can answer your questions on the build infrastructure, @sglyon

arnavs:
@sglyon Specifically to the CI, we're looking into a Docker-based solution. Something like https://docs.travis-ci.com/user/docker/? I think @jlperla wrote an issue on this, but I can't find it.

rst_files/wald_friedman.rst
arnavs: I think we can keep the cleverness, but maybe open up the terseness a little. Something like the following? Also fixes an issue where both subplots had the same metadata. 

```
dists = [Beta(1., 1.), Beta(3., 3.)] # Raw distributions. 
mix(dists, coef1) = MixtureModel(Beta, params.(dists), [coef1, 1-coef1]) # To generate our mixture models. 
mixes = mix.(Ref(dists), [0.25, 0.5, 0.75])

plot(
    plot(dists, labels = ["f_0", "f_1"], title = "Original Distributions"),
    plot(mixes, labels = ["1/4-3/4", "1/2-1/2", "3/4-1/4"], title = "Distribution Mixtures"), 
    ylab = "Density", ylim = (0, 2), layout = (2, 1) # Global settings across both plots
)
```

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-        curr_dist = p * f0 + (1 - p) * f1
-
-        # Get tomorrow's expected distribution through Bayes law
-        tp1_dist = clamp.((p * f0) ./ (p * f0 + (1 - p) * f1), 0, 1)
-
-        # Evaluate the expectation
-        EJ = dot(curr_dist, J.(tp1_dist))
-
-        return EJ
+    accept_x0(p, L0) = (one(p) - p) * L0
```
arnavs: I think these functions are excellent. Very close to the bare math.

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
 
-    function simulate_tdgp(wf, f; p0 = 0.5)
-        decision, p, t = simulate(wf, wf.f1; p0=p0)
+We can simulate an agent facing a problem and the outcome with the following function
 
-        correct = (decision == 1)
+.. code-block:: julia
 
-        return correct, p, t
+    @with_kw struct Problem
```
arnavs: Is there a reason we're not using a `NamedTuple`? Like: 

```
Problem = @with_kw (d0 = Beta(1,1), d1 = Beta(9,9), L0 = 2, L1 = 2, c = 0.2, p = 0.5, n = 100, return_output = false) # Constructor for problem objects. 
DefaultProblem = Problem()
ModifiedProblem = Problem(d0 = Beta(1, 3))
```

rst_files/wald_friedman.rst
jlperla: Yes, I never like clever.  Terse is not necessarily a virtue and don't forget point 1 of https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md#basic-principles

I think @arnavs has provided something which we can teach a little easier.

rst_files/wald_friedman.rst
jlperla: Yes: https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md#type-annotations-parameters-and-generic-programming

We want to avoid these kinds of structs at all costs.  Not to mention that this struct is not templatized for the distributions, so it would be holding abstract types.

Luckily, named tuples do it all.

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-            cdist[i] = correct
+    function simulation(problem)
+        @unpack d0, d1, L0, L1, c, p, n, return_output = problem
+        α, β = decision_rule(d0, d1, L0, L1, c)
+        outcomes = fill(false, n)
+        costs = fill(0.0, n)
+        trials = fill(0, n)
+        for trial in 1:n
+            truth = rand(1:2)
+            d = (d0, d1)[truth]
```
jlperla: This is very terse syntax.  I don't think that this is going to be easy for people to read.

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
+    accept_x1(p, L1) = p * L1
+    bayes_update(p, d0, d1) = p * pdf(d0, p) / pdf(MixtureModel([d0, d1], [p, one(p) - p]), p)
+    function draw_again(p, d0, d1, L0, L1, c, target)
+        candidate = 0.0
+        cost = 0.0
+        while candidate < target
+            p = bayes_update(p, d0, d1)
+            cost += c
+            candidate = min(accept_x0(p, L0), accept_x1(p, L1)) + cost
+            candidate ≥ target && break
```
jlperla: https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md#naming-conventions-comments-etc  ascii for control flow

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-        for (p_ind, p) in enumerate(pgrid)
-            # Payoff of choosing model 0
-            p_c_0 = payoff_choose_f0(wf, p)
-            p_c_1 = payoff_choose_f1(wf, p)
-            p_con = payoff_continue(wf, p, J_interp)
-
-            J_out[p_ind] = min(p_c_0, p_c_1, p_con)
+    function decision_rule(d0, d1, L0, L1, c)
+        function cost(p, d0, d1, L0, L1, c)
+            c ≥ zero(c) || throw(ArgumentError("Cost must be non-negative"))
```
jlperla: https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md#general-control-structures-and-code-organization Avoid the short-circuiting pattern when you just want to do an assert.

These is very tough to read for an introductory user.

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-            t = t + 1
-            p = bayes_update_k(wf, p, k)
-            if p < lb
-                decision = 1
-                break
-            elseif p > ub
-                decision = 0
-                break
+        for (left, right, root) in zip(reverse(left), reverse(right), reverse(roots))
+            if left ≠ 1 && right == 1
```
jlperla: Use ascii for control flow

rst_files/wald_friedman.rst
```diff
@@ -401,467 +345,201 @@ that appear on the right side of the Bellman equation as separate functions
...
-            k = rand(drv)
-            t = t + 1
-            p = bayes_update_k(wf, p, k)
-            if p < lb
-                decision = 1
-                break
-            elseif p > ub
-                decision = 0
-                break
+        for (left, right, root) in zip(reverse(left), reverse(right), reverse(roots))
```
jlperla: Ascii for control flow.  But I think that a comment could help a lot to explain what is going on here.

rst_files/wald_friedman.rst
Nosferican: Replace `&&` with `&`? Or `≠` with `!=`? `≠` is closer to the math.

Nosferican: Those the style also mean I should change `≥`/`≤` for control flow? Those are the other non-ASCII used in control flow in this lectures I believe.

rst_files/wald_friedman.rst
Nosferican: I strive to be, hehe. I took from the @arnavs version and terse it a bit.

> Run You Clever Boy, and Remember. - Clara Oswald

rst_files/wald_friedman.rst
Nosferican: Good catch. I have modified it to be consistent with that part of the style.

rst_files/wald_friedman.rst
Nosferican: Good catch. I have made the changes to use `NamedTuple`.

rst_files/wald_friedman.rst
Nosferican: Used some comments to help read the code better.

rst_files/wald_friedman.rst
jlperla: Yeah, I think that is right.  Though I really do love the unicode for ≠ if we are writing up a line of math where it is used for set inclusion, etc.

Nosferican: For set inclusion it would be `∈` or `∉`. I don't think `≠` is used outside control flow or tests... Very few exceptions is any. Probably best to discuss it later on the style guide, but I would cast my vote towards keeping control flow operations close to the math as well.

jlperla: Of course, you are right.  Then I don't think there is any reason to use ≠  except when doing DSL-y type stuff

Control flow is part of an algorithm, and not a direct implementation of a mathematical formula in the lecture notes.  Making the distinction between the direct implementation of the formulas from the math and the control flow of algorithms is a key goal.

Nosferican:
@arnavs Final check and and if all good, good to merge.

arnavs:
@Nosferican Sounds good. Will review and merge after class/meetings.

PR #159, opened 2018-10-06
Undeprecate AMSS

Nosferican:
Needs test and check of plot change. 0.7 only for now, until Dierckx.jl tags a 1.0 compat.


Nosferican:
@arnavs this is ready for review.

arnavs:
Looks good, we just need to add a `pkg"add Dierckx"` since we took that out of the shared QuantEcon environment.

jlperla:
Is `Dierckx.jl` really necessary here, or could Interpolations.jl do the job?  The missing feature in interpolations is splines with irregular grids, but when I did a quick look at this, it appeared the grid was regular?  If it appears feasible, maybe add in a new issue for someone to convert it.

Nosferican:
As an update... Been working on this long hours... The lecture "works" on 1.0, but is very dense and hard to follow. I have been able to make several passes improving it, but it needs at least some re-write to be "acceptable". I am also working on putting the tests before applying the re-write.

Nosferican:
That seems to be correct... Can't get the 0.6 version to run and tests were never performed on that one. It needs to be verified, but the lesson takes forever to run... Would be nice to have the Interpolations pass and merge it.

arnavs:
Merging this, thanks so much @Nosferican for wrangling with this lecture.

Issue #52, opened 2018-08-31
Fix Credits


arnavs:
We are moving to the crediting scheme in https://github.com/jstac/QuantEcon.lectures/issues/1195. 

Highlights: 

* Standardized format for coauthors at the top of each page. (when applicable)
* Removing all "inline" citations/attributions (e.g., in comments, cells, Markdown, etc.) and moving these to a global credits page. There should be no names in the code. 
* Credits page will be in a different repo. Details on where to submit them TBD. 

No need to duplicate the whole checklist here. Can just be mindful of this as we proceed with the rewrite, and then do a final pass to make sure we got them all. 

cc: @jstac, @jlperla, @Nosferican

jlperla:
Sounds good. For now, if there is someone not listed in https://lectures.quantecon.org/about_lectures.html#id7 I think you should add a note to this issue so we don't lose them

Nosferican:
Probably best to do all in one repo after the first pass.

arnavs:
OK, per @jlperla, I did a first pass; below are the names of people whose names come up in the lectures but who aren't credited on the global attribution paragraph. Some of these cites are for `QuantEcon.jl` contributions which are tightly bound to the lectures. 

* Max Huber 
* Balint Szoke 
* Victoria Gregory 
* Willem Hekman 
* Guanlong Ren (with Will, for solving Ex. 1 in `linear_algebra`)
* Alberto Polo (for editing the `Wald_Friedman_jl.ipynb`
* Ranjan Ananthamaran (for working on `lss.jl` in QuantEcon, which powers the `linear_models` lecture). 
* Zac Cranko (for the `lqcontrol` file in the QuantEcon package). 
* Colin Beckingham (for work on the `QuantEcon.jl/src/markov/markov_approx.jl` file, which solves Exercise 3). 

Also, below is the table I was using to keep track of references to people by name. Doesn't directly apply to a task, but I figure it's useless to just destroy it. Some lines may omit frequent contributors.

Lecture Filename | Author(s) | Contributors | 
----------------- | ---------  | ------------
`additive_functionals` | Chase Coleman and Balint Szoke (joint) | Shunsuke Hori | 
`aiyagari` | N/A | N/A |
`amss` | N/A | N/A | 
`arellano` | N/A | Chase Coleman |
`arma` | N/A | N/A | 
`career` | N/A | N/A | 
`classical_filtering` | N/A | Shunsuke Hori | 
[...]

PR #161, opened 2018-10-06
Update Linear Algebra

Nosferican:
Not much of a re-write
- Fixed credits
- Dropped LaTeXStrings
- Fixed activate
- Fixed alignment
Open to suggestions.


rst_files/linear_algebra.rst
```diff
@@ -81,21 +76,20 @@ the point
 
 The following figure represents three vectors in this manner
 
+Activate the project environment, ensuring that ``Project.toml`` and ``Manifest.toml`` are in the same location as your notebook
+
 .. code-block:: julia
-    :class: test
 
-    using Test # Should put this near the top of every notebook.
+    using Pkg; Pkg.activate(@__DIR__); #activate environment in the notebook's location
```
jlperla: We can't have the activate stuff inside of the test block, otherwise it won't run when we generate the notebooks without tests.

But perhaps we need to have the activate in its own block before the `using Test` block?

Nosferican: That's not a test block. The activate cell was after code so I moved it to the top before any code.

jlperla: Sorry, looks like I misread the diff

rst_files/linear_algebra.rst
```diff
@@ -11,6 +11,8 @@ Linear Algebra
 
 .. contents:: :depth: 2
 
+Co-authors: `Spencer Lyon <https://github.com/sglyon>`__ and `Victoria Gregory <http://github.com/vgregory757>`__.
```
jlperla: @Nosferican  Based on the https://github.com/QuantEcon/lecture-source-jl/issues/52 I think the intention was that we would ensure these were in the contributors, not coauthors of the lectures.  That was the common plan for all lectures with code contribution.  @jstac can verify.

Nosferican:
@arnavs Ready for review.

rst_files/linear_algebra.rst
jlperla: Just a reminder to remove this entirely.  See that #52  for details, but basically you do not need to add in coauthors.

jlperla:
@Nosferican See comment about the credits links, etc.

rst_files/linear_algebra.rst
```diff
@@ -992,10 +992,12 @@ As expected, the image :math:`Av` of each :math:`v` is just a scaled version of
     A = [1 2
          2 1]
     evals, evecs = eigen(A)
+    
     a1, a2 = evals
     evecs = selectdim.(Ref(evecs), 1, 1:2)
     eig_1 = reduce(hcat, ([0, v] for v in first.(evecs)))
```
jlperla: I think this pattern of a reduction on `hcat` is **way** to difficult to understand for new students.  It took me a while to figure it out.

We have never even taught them that `hcat` is a function, that reduce is a higher order function, etc. 

The old for loops may be easier to understand.

rst_files/linear_algebra.rst
```diff
@@ -92,22 +96,16 @@ Activate the project environment, ensuring that ``Project.toml`` and ``Manifest.
...
-    labels = []
 
     # Create matrices of x and y values, labels for plotting
-    for i ∈ eachindex(vecs)
-        v = vecs[i]
-        x_vals[2, i] = v[1]
-        y_vals[2, i] = v[2]
-        labels = [labels; (1.1 * v[1], 1.1 * v[2], "$v")]
-    end
+    x_vals = reduce(hcat, ([0, v] for v in first.(vecs)))
```
jlperla: See other comments.  Is there any way to make the `x_vals` and `y_vals` constructions easier to understand?  If not, I would rather stick with the for loop in this case.

rst_files/linear_algebra.rst
```diff
@@ -183,41 +181,26 @@ Scalar multiplication is illustrated in the next figure
...
-        x_vals[2, i] = v[1]
-        y_vals[2, i] = v[2]
-        labels = [labels; (v[1] + 0.4, v[2] - 0.2, string(s, "x"))]
-    end
+    x = [2, 2, 2]
+    scalars = [-2, 1, 2]
+    vals = x .* scalars
+    x_vals = reduce(hcat, ([0, v] for v in vals))
+    y_vals = x_vals
+    labels = map((v, s) -> (v + 0.4, v - 0.2, string(isone(s) ? "" : s, "x")), vals, scalars)
```
jlperla: This is pretty tricky code for what we really need.  We can certainly use the `map`, as we taught it, but I don't think this is really worth it for this example.  Why not hardcode the 3 labels?

rst_files/linear_algebra.rst
```diff
@@ -336,55 +314,29 @@ The span is a 2 dimensional plane passing through these two points and the origi
...
-    xr2 = range(x_min, stop = x_max, length = grid_size)
-    yr2 = range(y_min, stop = y_max, length = grid_size)
+    xr2 = range(x_min, x_max, length = grid_size)
+    yr2 = range(y_min, y_max, length = grid_size)
     z2 = zeros(grid_size, grid_size)
-    for i ∈ 1:grid_size
-        for j ∈ 1:grid_size
-            z2[j, i] = f(xr2[i], yr2[j])
-        end
+    for (position, value) in zip(eachindex(z2), (f(x, y) for x in xr2 for y in yr2))
```
jlperla: I have no idea what this code does.  Lets go back to some variation of the for loop, which is crystal clear.

rst_files/linear_algebra.rst
Nosferican: The for loops were pretty hard to understand as well. Some alternatives are
```
reduce(hcat, vcat.(0, first.(evecs)))
```
```
eig_1 = [zero(n, 1), first.(evecs)']
```

rst_files/linear_algebra.rst
```diff
@@ -1099,27 +989,18 @@ As expected, the image :math:`Av` of each :math:`v` is just a scaled version of
...
-
-    for i ∈ eachindex(evecs)
-        v = evecs[i]
-        eig_1[2, i] = v[1]
-        eig_2[2, i] = v[2]
-    end
-
-    x = range(-5, stop = 5, length = 10)
+    a1, a2 = evals
+    evecs = selectdim.(Ref(evecs), 1, 1:2)
```
jlperla: Is this complexity really necessary?   I am not sure what this code is doing.  Why not just use a simple slice?

rst_files/linear_algebra.rst
Nosferican: We could hardcode the labels. Is a matter of whether is worthwhile for them to see how to generate the labels programatically.

rst_files/linear_algebra.rst
Nosferican: The slice version is
```
[ evecs[i,:] for i in 1:size(evecs, 1) ]
```
If that is more clear I can do that one.

rst_files/linear_algebra.rst
Nosferican: That's the Pythonic `[ f(x,y) for x in x_vals for y in y_vals ]`. It is the same as
```
reshape([ f(x,y) for x in x_vals for y in y_vals ], grid_size, grid_size)
```
I can go back to the nested for loops.

rst_files/linear_algebra.rst
jlperla: This is too complicated for intro users

rst_files/linear_algebra.rst
jlperla: That looks pretty clear to me, or something along those lines.

rst_files/linear_algebra.rst
jlperla:
```
eig_1 = [zero(n, 1), first.(evecs)']
```
That is still pretty tricky.  Do we really need to use broadcasting here?  This is one of the first lectures people are going to read after just barely learning julia.  How big is the vector?  Can we just concatenate it with a hardcoded number of columns?

rst_files/linear_algebra.rst
Nosferican: I will just hard-code all the values at least for the ones in this section.

rst_files/linear_algebra.rst
jlperla:
```
  for j ∈ 1:grid_size
            z2[j, i] = f(xr2[i], yr2[j])
        end
```
or something like that just seems so easy to mentally parse here.  `reshape` is a tough concept for people

Nosferican: That will just give a `Vector` and not the `Matrix`, hence the `reshape`.
The alternative is what I had of having an iterator with all the values in order and fill the `Matrix` one by one in the efficient order (relying on `eachindex(::Array)`).

jlperla: Check this out!
```
[ f(x,y) for x in x_vals, y in y_vals ]
```
Much better than a for loop in this case.

jlperla: (note the change of the `for` to a comma)?  I think this works?

arnavs: @jlperla This works, but it does a Cartesian grid instead of a pairwise thing. 

```
julia> [(a, b) for a in 1:5, b in 1:5]
5×5 Array{Tuple{Int64,Int64},2}:
 (1, 1)  (1, 2)  (1, 3)  (1, 4)  (1, 5)
 (2, 1)  (2, 2)  (2, 3)  (2, 4)  (2, 5)
 (3, 1)  (3, 2)  (3, 3)  (3, 4)  (3, 5)
 (4, 1)  (4, 2)  (4, 3)  (4, 4)  (4, 5)
 (5, 1)  (5, 2)  (5, 3)  (5, 4)  (5, 5)
```

jlperla: That is because `(a, b)` returns a tuple.  try it with `a + b` which returns an integer.

arnavs: Same story:

```
julia> [a+b for a in 1:5, b in 1:5]
5×5 Array{Int64,2}:
 2  3  4  5   6
 3  4  5  6   7
 4  5  6  7   8
 5  6  7  8   9
 6  7  8  9  10
```

It actually looks like a Cartesian grid is what we want in this case, unless I'm mistaken. But good to know for this pattern in the future.

jlperla: I am not sure what you mean by pairwise, but I guess it depends on what that `for` loop was doing.  I was responding on the fact that you couldn't get a matrix from a comprehension, which you can.

Regarldess, the pattern may be useful, as might taking the transpose of it.  This is a perfect example of where a few words on the algorithm explaining what we want and what the code needs to do could help.

rst_files/linear_algebra.rst
Nosferican: The version corresponds to the changed version in master. If the credits pass hasn't been done yet, I think is better to do all of those in one pass rather than in each lecture PR.

arnavs: I think the credits pass was done more than 14 days ago (i.e., if you go to the current master, the names were taken out).

Nosferican: Seems like I resolved the conflicts in between having it merged. Will try to make a clean up to make sure it is up to date.

arnavs: No problem.

rst_files/linear_algebra.rst
```diff
@@ -91,42 +78,41 @@ the point
 
 The following figure represents three vectors in this manner
 
+Activate the project environment, ensuring that ``Project.toml`` and ``Manifest.toml`` are in the same location as your notebook
+
+Setup
+------------------
+
+Activate the ``QuantEconLecturePackages`` project environment and package versions
```
arnavs: Should be changed to reflect latest `master`. Same with the other PR.

rst_files/linear_algebra.rst
```diff
@@ -807,66 +704,25 @@ If so, is it always unique?
...
-        foreground_color_border = :white)
-
-    ybar = 2.6
-    plot!(x, x .* 0 .+ ybar, color = :black, linestyle = :dash, legend = :none)
-    annotate!(0.04, 0.91 * ybar, L"$y$")
-
-    plot(p1, p2, layout = (2, 1), size = (600, 700))
+    :class: collapse
+
+    y_min, y_max = extrema(0.6cos.(4x) + 1.3 for x in range(-2, 2, length = 100))
```
arnavs: This is still a bit confusing, and also the broadcasting is redundant. What I'd suggest is: 

```
vals = [0.6cos(4x) + 1.3 for x in range(-2, 2, length = 100)] # Now it's a list comprehension, which they know.
y_min, y_max = extrema(vals)
```

rst_files/linear_algebra.rst
```diff
@@ -807,66 +704,25 @@ If so, is it always unique?
...
+    plt1 = plot(x -> 0.6cos(4x) + 1.3, xlim = (-2, 2), label = "f")
```
arnavs: We could then replace this with `plot(vals)`.

rst_files/linear_algebra.rst
```diff
@@ -1109,39 +956,29 @@ The next figure shows two eigenvectors (blue arrows) and their images under :mat
...
-    labels = []
-
-    for i ∈ eachindex(evecs)
-        v = evecs[i]
-        eig_1[2, i] = v[1]
-        eig_2[2, i] = v[2]
-    end
 
+    a1, a2 = evals
+    eig_1 = [0 0; evecs[:,1]']
```
arnavs: Bit terse, but I don't think this is above the students' level. Maybe worth reminding them of what `evecs` returns.

rst_files/linear_algebra.rst
Nosferican: The whole point is to avoid allocating the values (especially for the later lectures when there are quite substantial). I am fixing the legacy broadcast.

rst_files/linear_algebra.rst
Nosferican: The nice aspect of `plot` is that it takes a function and one needs not to specify a grid.

rst_files/linear_algebra.rst
Nosferican: There is a markdown cell just below doing exactly that.

rst_files/linear_algebra.rst
jlperla: True, and I think we should be using that frequently.

If arnav thought this was tough to read, maybe having this as a separate function would be better
```
f(x) = 0.6cos.(4x) + 1.3 f
y_min, y_max = extrema( f.(range(-2, 2, length = 100))
plt1 = plot(f, xlim = (-2, 2), label = "f")
```
Or something like that.

arnavs: After thinking about it some more, I don't think I have an issue with this. I was mainly saying that if we switched the thing above to `vals`, then we might as well switch this. But I like the function feature.

rst_files/linear_algebra.rst
arnavs: Gotcha. I think the issue here with `extrema(0.6cos.(4x) + 1.3 for x in range(-2, 2, length = 100))` is that it's not clear exactly what I'm giving to `extrema`. The function takes an `AbstractArray`, but I can't just put the argument in the REPL and get a collected array or range or something.

@jlperla, thoughts?

Nosferican:
```
extrema(itr) -> Tuple # An iterable not an AbstractArray
```
It looks like
```
extrema(f(x) for x in range(-2, 2, length = 100))
```

jlperla:
What is wrong with:

`extrema(f.(range(-2.0, 2.0, length=100))`


Seems more clear to me?

Nosferican:
The broadcast version materializes the whole array needlessly.

rst_files/linear_algebra.rst
arnavs: Right, but `f(x) for x in range(-2, 2, length = 100)` isn't an iterable (that is, I couldn't write `iterobj = x^2 for x in range(-2, 2, length = 100)` or something).

arnavs:
OK. I guess we can keep this for now, and tell them that `()` will give them an uncollected list comprehension if someone asks.

jlperla:
Guys:  what is the goal here?  **Clarity**.

Comprehensions can be useful for clarity, or they can make it far more complicated than it needs to be.  

```
julia> using BenchmarkTools

julia>  extrema(0.6cos.(4x) + 1.3 for x in range(-2, 2, length = 100))
(0.700029536547185, 1.898885395947853)

julia> @btime extrema(0.6cos.(4x) + 1.3 for x in range(-2, 2, length = 100))
  2.122 μs (0 allocations: 0 bytes)
(0.700029536547185, 1.898885395947853)

julia> f(x) = 0.6* cos(4x) + 1.3
f (generic function with 1 method)

julia> x = range(-2, 2, length=100)
-2.0:0.04040404040404041:2.0

julia> @btime extrema(f.($x))
  2.389 μs (1 allocation: 896 bytes)
(0.700029536547185, 1.898885395947853)
```

.... so all of that lack of clarity has given you 0.2μs in speed, and 896 bytes.

Not worth it at all.

arnavs:
OK, will change it back.

Issue #67, opened 2018-09-06
PyPlot and Gadfly to Plots


Nosferican:
Every plot should use the `Plots.jl` API with the GR backend. In case of a plot not being able to be generated with the Plots API (bring the special case to attention for discussion).

- [ ] [Job Search IV: Modeling Career Choice](https://lectures.quantecon.org/jl/career.html)
- [x] [Optimal Growth I: The Stochastic Optimal Growth Model](https://lectures.quantecon.org/jl/optgrowth.html) (#84)
- [x] [Optimal Growth III: The Endogenous Grid Method](https://lectures.quantecon.org/jl/egm_policy_iter.html) (#95)
- [x] [Optimal Savings II: LQ Techniques](https://lectures.quantecon.org/jl/perm_income_cons.html) (#104)
- [x] [Consumption and Tax Smoothing with Complete and Incomplete Markets](https://lectures.quantecon.org/jl/smoothing.html) (#97)
- [x] [A Lake Model of Employment and Unemployment](https://lectures.quantecon.org/jl/lake_model.html) (#103)
- [x] [Rational Expectations Equilibrium](https://lectures.quantecon.org/jl/rational_expectations.html)
- [x] [Markov Perfect Equilibrium](https://lectures.quantecon.org/jl/markov_perf.html) (#102)
- [x] [Uncertainty Traps](https://lectures.quantecon.org/jl/uncertainty_traps.html)
- [x] [Default Risk and Income Fluctuations](https://lectures.quantecon.org/jl/arellano.html) (Gadfly|priority)
- [x] [Globalization and Cycles](https://lectures.quantecon.org/jl/matsuyama.html)
- [x] [Covariance Stationary Processes](https://lectures.quantecon.org/jl/arma.html) (#126)
- [x] [Estimation of Spectra](https://lectures.quantecon.org/jl/estspec.html)
- [x] [Additive Functionals](https://lectures.quantecon.org/jl/additive_functionals.html)
- [x] [Multiplicative Functionals](https://lectures.quantecon.org/jl/multiplicative_functionals.html)
- [x] [Classical Control with Linear Algebra](https://lectures.quantecon.org/jl/lu_tricks.html) (#122)
- [x] [Optimal Taxation in an LQ Economy](https://lectures.quantecon.org/jl/lqramsey.html) (#120)
- [x] [Optimal Taxation with State-Contingent Debt](https://lectures.quantecon.org/jl/opt_tax_recur.html) 
[...]

PR #163, opened 2018-10-08
lucas_model.rst Interpolations and Plots

ajozefiak:
I noticed that this lecture was also using PyPlot (and was not listed in the Plots issue #67) so I quickly converted the PyPlot to Plots.


arnavs:
Perfec.

Issue #146, opened 2018-10-02
Interpolations pass


jlperla:
Change everything to use Interpolations.jl and remove references to the QuantEcon

arnavs:
Checklist for the `LinInterp`:

- [ ] Wald-Friedman
- [ ] Robustness 
- [ ] Optimal Growth 
- [ ] `opt_tax_recur`
- [ ] `odu`
- [ ] Lucas Model 
- [ ] Job V 
- [ ] `egm_policy_iter`
- [ ] `coleman_policy_iter`
- [ ] `amss` *

PR #164, opened 2018-10-08
Egm policy iter interp

ajozefiak:
#146


arnavs:
Great. And the increase in accuracy from `O(10^-7)` to `O(10^-16)` is a great example for why we should use packages for this stuff.

PR #169, opened 2018-10-16
Adam's Matsuyama PR

arnavs:
Need to convert the last one.


arnavs:
@Nosferican do you know where`withfig` comes from here? Can't find a definition for it.

PR #173, opened 2018-10-18
Rewrite "Finite Markov Chains"

arnavs:
#160. Nothing major. Just mainly: 

1. Moved away from QuantEcon.jl's `DiscreteRV` to Bernoulli or Categorial from Distribution.
2. Made some style guide/legibility changes.

### DIFF
--- a/rst_files/finite_markov.rst
+++ b/rst_files/finite_markov.rst
@@ -34,14 +34,14 @@ Setup
 
 Activate the ``QuantEconLecturePackages`` project environment and package versions
 
-.. code-block:: julia 
+.. code-block:: julia
 
     using InstantiateFromURL
     activate_github("QuantEcon/QuantEconLecturePackages");
 
-.. code-block:: julia     
-    
-    # Import packages. 
+.. code-block:: julia
+
+    # Import packages.
     using LinearAlgebra, Statistics, Compat
     using QuantEcon, Random, Plots, Printf, Distributions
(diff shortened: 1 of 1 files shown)


Nosferican:
Not sure if we want to tackle this at the moment, but in my experience some of the images should be reduced from 100% size. If that is the case in your end, you could do so now or leave it to the images pass. Other than that, it seems fine.

arnavs:
Sure, and we can add that to the style guide as well (@jlperla). Something like "make sure your images take up no more than X% of screen width."

arnavs:
Sounds good @jlperla. Going to merge this then and we can revisit towards the end.

jlperla:
Yeah, this doesn't seem to use anything too advanced that we didn't introduce in the intro slides.

rst_files/finite_markov.rst
```diff
@@ -1428,28 +1268,21 @@ compare it to the stationary probability.
...
 
+    α = 0.1 # probability of getting hired
+    β = 0.1 # probability of getting fired
+    N = 10000
+    p_bar = β / (α + β) # steady-state probabilities
+    P = [1 - α   α
+         β   1 - β] # stochastic matrix
     mc = MarkovChain(P)
+    labels = ["start unemployed", "start employed"]
+    y_vals = Array{Vector}(undef, 2) # sample paths holder
```
arnavs: Should definitely replace this with a matrix if possible.

PR #176, opened 2018-10-20
Global Setup

### LINKED ISSUE #175: Make the set up lines a static file
Similar to the `lecture_howto_jl.raw`, the set up is repeated across each lecture. I think it would be a good candidate for a static file.

arnavs:
Resolves #175. In particular: 

1. Replace all the `using InstantiateFromURL; activate_github("..."); using ...` code blocks with a literal include. 

2. Adds a line to the makefile which populates that static file with a tag. The default is `v0.3.1`. You can change it either in the Makefile, or by passing a command-line argument to `make`.


arnavs:
Note: This won't work on macOS, unless your `/bin/bash` is using `gnu-sed` instead of the FreeBSD (that is, even if your interactive shell path points there, a script won't necessarily).

jlperla:
What do you need me to look at?

arnavs:
@Nosferican, OK, sure.

jlperla:
The definetely should be in separate cells, so you can do the `using` without running the instantiation.

I am not entirely sure if we even want the `using LinearAlgebra, Statistics, Compat` generated within this static include at all?

If we are putting most of the package dependencies at the top of the file, it would mean that we would have different sets of `using`...

Nosferican:
The issue is that one cannot call `using ...` without the `activate_github` since that is what `activate` the environment. Also makes the lectures error out currently if offline.

arnavs:
I don't think the instantiation is a big issue, because unless users delete packages, it should only happen once. And if they use a tag like `v0.3.1`, we won't even download new resources.

jlperla:
Maybe I misunderstand, but I don't think that is true
- If you choose not to execute the `activate_github` cell, it will just use your `(v1.0)` environment, which you would have to have built up.
- If you have already loaded the `v0.3.1` set of notebooks, for example, it doesn't go back online unless you go `force = true`

But these are all of the reasons to separate out the instantiation from the `using` in separate cells.

arnavs:
Right, if you choose not to use the `QuantEcon/QuantEconLecturePackages`, then it'll use the default one. If we want to target that, I agree cleanest is just to strike the `using ...` from the setup script.

Nosferican:
Without the tag it is online only since it will always check with master and fail. With the tag it now works fine offline.

jlperla:
Correct.  Even when https://github.com/QuantEcon/InstantiateFromURL.jl/issues/21 is complete (which ensures that the `instantite;precompile` is not done each done) it still needs to go online to check if the SHA changed.  Since we are putting in the tags for each notebook right now, there hasn't been much need to finish that issue.

Nosferican:
I am getting,
```
[user]:QuantEcon_lecture-source-jl [user]$ make jupyter
sed -i 's/DEPVER/v0.3.1/g' rst_files/_static/includes/deps.jl 
sed: -i may not be used with stdin
make: *** [jupyter] Error 1
```

jlperla:
Is this `sed` thing worth the trouble?  Why not just hardcode it in the `deps.jl`?  We can bump the version there when we need to, and shouldn't be changing it all that often.

arnavs:
@Nosferican, please read the README or the comment above about using this on macOS. The issue is that `sed` on Mac is the FreeBSD version, which requires an additional `sed -i '' ....`. 

@jlperla, I think it makes sense for the Linux. But we can drop it if most people aren't building this on Linux.

Nosferican:
I saw the comment, but I don't think the Mac OS specific issue made it to the README (it only has the Windows specific issues).

Issue #194, opened 2018-10-23
Organization of github PRs for new lectures


jlperla:
I need to start having new lectures and major rewrites as PRs so we don't have half-finished or orphaned material when generating HTML.  New lectures can be setup as a WIP with the new file and the changes in the accompanying index.

A few of the things to do:
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/index_getting_started_julia.rst
  - That is wrong.  It should end with `introduction_to_types`  The rest are in the "more" section
  - I don't anticipate any more lectures in there.
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/index_more_julia.rst
  - This is where the additional lectures material will go
  - Leave the `generic_programming` in there, since I will finish it today and work off of master.
  - Otherwise, only the `need_for_speed` is sufficiently coherent and updated to be generated for the html.
  - I think the others should all be removed, where the addition to the index there is part of the PR for the lecture itself.

Which leads to the high priority additional lectures
- [x]  PR for https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/version_control.rst where we will put the Git tutorials, with accompanying change to the index files
- [x] same for https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/testing.rst where we will put in unit testing
- [x] for https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/tools_editors.rst where we will put the material Atom and the REPL, creating project files/etc.
- [x] PR for https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/more_types.rst  with accompanying change to the index files

And the lower priority ones, which we should have a PR for but are unlikely to be finished off anytime soon
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/more_types.rst  
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/plots.rst for updates on the plotting ecosystem
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/file_io.rst for file/io, using bigdata type files, loading R datasets, etc.
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/packages.rst for additional packages in the julia ecosystem
- [x] https://github.com/QuantEcon/lecture-source-jl/blob/master/rst_files/need_for_speed.rst a rewrite of this for the newest Julia.

[...]

PR #195, opened 2018-10-23
"Version Control" Lecture

arnavs:
First pass. Creating a PR per @jlperla in #194. Will delete this from `master` as well


jlperla:
GitHub Desktop + GitHub Website for dummies.
- Only things they really need to do as beginners collaborating on a project
- At the end, can show screenshots of simple PR for an open-source project

jlperla:
For screenshots:  As you go through, take many more that you think you will really need, because it will be hard to fill in the blanks otherwise if we decide things are unclear:
* In the early parts, Expectations are a good choice, but nothing about PRs (until the Open Source section)
* Creating a new example private repo called `ExampleProject` or something like that
* We can also use the lectures repo as well for screen shots, though there is a lot of noise.

arnavs:
@jlperla I think this is about ready for you to review, modulo stuff like image sizes. It could be that I went overkill on the screenshots as you suggested.

jlperla:
A few modifications/fixes to do:
Not sure what went wrong, but can't see the screenshots.   Maybe move them up from the `/git` subdirectory for consistency with the other static assets... whether it fixes it or not.

Why don't add in a simple "git commandline basics" for people, since this stuff comes up?  My suggestion would be just the following:
- [ ] Explain how to get the Git Bash in windows (I am not sure a screenshot is necessary, but you could
- [ ] Tell them they they can use `cd` and `ls`, and simple linux commands
- [ ] Tell them how to clone, maybe with some example repo? (preferably not something attached to one of our accounts, but that is OK if necessary)  I bet there is some standard git one they use for the docs?
- [ ] Tell them how to pull, i.e. `cd` into the folder, then `git pull`
- [ ] Tell them how to do the hard reset, i.e. the `git fetch origin; git reset --hard HEAD; git pull`
- [ ] I think we should avoid telling them how to commit/add/etc., but you have those links to give them.

jlperla:
I think we should avoid telling them to clone manually any Julia package (e.g. Expectations.jl) in case they get confused.  One thing we could do is add in an example github repo on the QuantEcon organization?

I would prefer a name less generic than `git-setup`.  Maybe we could do the workflow with one called `QuantEcon/example_repository`?  Would it be hard to redo the screenshots where they are not connected to your personal account?  I think it would look more professional when the v1.0 of QE is released, and would allow us to keep those repos permanent?  Doing the fork step to the ubcecon is OK, since we need at least one other org/account to do it.

arnavs:
Modifications sound great. Will flatten the figures directory (although we should eventually pin down why it works on local machines but fails on the syzygy).

As for the screenshots, it took a while to make them, but I see your point. If I could get repo access to QuantEcon, then maybe over the weekend I could reproduce them.

jlperla:
After we redo the screenshots, maybe it makes sense to show a screenshot of the github issues link: https://github.com/issues for the test user, where we have shown them post and assigned an issue... or something like that.

arnavs:
@jlperla Put in a few words about the command line under the "additional resources/troubleshooting." The only other command I added was `git clean -fd` to remove untracked files and dirs.

arnavs:
@jlperla Changes look good to me. Merging now.

PR #202, opened 2018-10-23
"File IO" Lecture

arnavs:
Not done yet.


jlperla:
We should add eventually add details on:
- writing cross-platform code with files (e.g. `joinpath`, etc.)
- downloading files from the web
- Pandas/etc. loading of data
- Storing as HDF files for matlab/python
- RDatasets for exploration
- Maybe DataDeps when it is ready.

PR #204, opened 2018-10-23
"Need for Speed" Lecture Rewrite

arnavs:
This is currently level with master (modulo an empty commit `4523101`). Can add to the branch when we get around to this.


jlperla:
@Nosferican If it is easier, we could nuke this PR and just add a new one?

Nosferican:
Guess rebasing does the same, but the diff does get messy. Sure, let's do that.

PR #205, opened 2018-10-23
Rewrite "A First Look at the Kalman Filter"

arnavs:
Still WIP. Changes include moving away from functions that plot via side-effects (i.e., `plot_gen_gaussian()`), using `Distributions.jl` instead of a hand-rolled PDF, etc.

### DIFF
--- a/rst_files/kalman.rst
+++ b/rst_files/kalman.rst
@@ -104,54 +104,6 @@ This density :math:`p(x)` is shown below as a contour map, with the center of th
   using Plots, LaTeXStrings, LinearAlgebra, Compat
     gr(fmt=:png)
 
-  function bivariate_normal(X,
-                            Y,
-                            σ_x = 1.0,
-                            σ_y = 1.0,
-                            μ_x = 0.0,
-                            μ_y = 0.0,
-                            σ_xy = 0.0)
-      Xμ = X .- μ_x
-      Yμ = Y .- μ_y
-
-      ρ = σ_xy / (σ_x * σ_y)
-      z = Xμ.^2 / σ_x^2 + Yμ.^2 / σ_y^2 - 2 * ρ .* Xμ .* Yμ / (σ_x * σ_y)
-      denom = 2π * σ_x * σ_y * sqrt(1 - ρ^2)
-      return exp.(-z / (2 * (1 - ρ^2))) ./ denom
-  end
-
-
-  # == Set up the Gaussian prior density p == #
-  Σ = [0.4  0.3
-       0.3 0.45]
-  x_hat = [ 0.2
-           -0.2]''
-
-  # == Define the matrices G and R from the equation y = G x + N(0, R) == #
-  G = I
-  R = 0.5 .* Σ
-
-  # == The matrices A and Q == #
-  A = [1.2    0
-       0   -0.2]
-  Q = 0.3 .* Σ
-
-  y = [2.3, -1.9]
-
-  # set up grid for plotting 
-  x_grid = range(-1.5, 2.9, length = 100) 
-  y_grid = range(-3.1, 1.7, length = 100)
-  X = repeat(x_grid', length(y_grid), 1)
-  Y = repeat(y_grid, 1, length(y_grid))
-
-  function gen_gaussian_plot_vals(μ, C)
-      "Z values for plotting the bivariate Gaussian N(μ, C)"
-      m_x, m_y = μ[1], μ[2]
-      s_x, s_y = sqrt(C[1, 1]), sqrt(C[2, 2])
-      s_xy = C[1, 2]
-      return bivariate_normal(X, Y, s_x, s_y, m_x, m_y, s_xy)
-  end
-
     # set up prior objects 
     Σ = [sqrt(0.4)  0.3
          0.3 sqrt(0.45)]
--- a/rst_files/kalman.rst
+++ b/rst_files/kalman.rst
@@ -101,7 +101,7 @@ This density :math:`p(x)` is shown below as a contour map, with the center of th
 .. code-block:: julia
   :class: collapse
 
-  using Plots, LaTeXStrings, LinearAlgebra, Compat
+  using Plots, LaTeXStrings, LinearAlgebra, Compat, Distributions
     gr(fmt=:png)
 
     # set up prior objects 
@@ -225,7 +225,7 @@ The original density is left in as contour lines for comparison
     Σ_F = Σ - M * G * Σ
 
     # plot the new density on the old plot 
-    newdist = MvNormal(x_hat_F, Σ_F)
+    newdist = MvNormal(x_hat_F, Symmetric(Σ_F)) # because Σ_F
     contour!(x_grid, y_grid, (x, y) -> pdf(newdist, [x, y]), fill = false, levels = 6, color = :grays, cbar = false)
 
 .. code-block:: julia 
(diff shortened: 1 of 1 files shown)


arnavs:
There's a [bug](https://discourse.julialang.org/t/multivariate-normal-with-positive-semi-definite-covariance-matrix/3029/12) in Distributions.jl which prevents us from proceeding further. But otherwise, we can basically use the procedure for the first plot on all of them.

arnavs:
Ready for review.

rst_files/kalman.rst
```diff
@@ -101,71 +101,42 @@ This density :math:`p(x)` is shown below as a contour map, with the center of th
 .. code-block:: julia
   :class: collapse
 
-  using Plots, LaTeXStrings, LinearAlgebra
+  using Plots, LaTeXStrings, LinearAlgebra, Compat, Distributions
```
Nosferican: We shouldn't be using LaTeXStrings (Compat should have already been loaded)

rst_files/kalman.rst
```diff
@@ -101,71 +101,42 @@ This density :math:`p(x)` is shown below as a contour map, with the center of th
...
     gr(fmt=:png)
```
Nosferican: The indentation of this whole cell is off. Using 4 spaces.

rst_files/kalman.rst
Nosferican: Divide the `using` and `gr` statements from the other content on this cell.

rst_files/kalman.rst
```diff
@@ -666,29 +619,22 @@ Exercise 1
...
-    # == Run == #
 
-    N = 5
     xgrid = range(θ - 5, θ + 2, length = 200)
-    densities = []
-    labels = []
-    for i ∈ 1:N
+    densities = Array{Array{Float64, 1}, 1}(undef, 5) # we'll do 5 rounds of updating
+    labels = ["t=1", "t=2", "t=3", "t=4", "t=5"]
+    for i ∈ 1:5
```
Nosferican: replace with `in`

rst_files/kalman.rst
```diff
@@ -666,29 +619,22 @@ Exercise 1
...
     kalman = Kalman(A, G, Q, R)
     set_state!(kalman, x_hat_0, Σ_0)
-    # == Run == #
 
-    N = 5
     xgrid = range(θ - 5, θ + 2, length = 200)
-    densities = []
-    labels = []
-    for i ∈ 1:N
+    densities = Array{Array{Float64, 1}, 1}(undef, 5) # we'll do 5 rounds of updating
```
Nosferican: Style no concrete.

rst_files/kalman.rst
```diff
@@ -666,29 +619,22 @@ Exercise 1
...
+    labels = ["t=1", "t=2", "t=3", "t=4", "t=5"]
+    for i ∈ 1:5 
         # Record the current predicted mean and variance, and plot their densities
         m, v = kalman.cur_x_hat, kalman.cur_sigma
-        push!(densities, pdf.(Normal(m, sqrt(v)), xgrid))
-        push!(labels, LaTeXString("\$t=$i\$"))
+        densities[i] = pdf.(Normal(m, sqrt(v)), xgrid)
```
Nosferican: Instead of a pre-allocation and loop, why not a list comprehension.

rst_files/kalman.rst
```diff
@@ -713,34 +659,34 @@ Exercise 2
...
     ϵ = 0.1
     kalman = Kalman(A, G, Q, R)
     set_state!(kalman, x_hat_0, Σ_0)
 
     nodes, weights = qnwlege(21, θ-ϵ, θ+ϵ)
 
-    T = 600
+    T = 300
     z = zeros(T)
     for t ∈ 1:T
```
Nosferican: Same here (`in`)

Nosferican:
Some plots still have unicode which shows as `?`.

Nosferican:
Alignment issues such as
```
Σ = [0.9 0.3
            0.3 0.9]
```

Nosferican:
Can we replace the images with Julia code?

rst_files/kalman.rst
arnavs: Took out `LaTeXStrings` and `Compat`. Same is true for `LinearAlgebra` as well.

rst_files/kalman.rst
arnavs: Thanks. I think it's hard to enforce indentation and the like cross-platform, so as long as it doesn't break the rST (that is, unless indentation _levels_ are off), I'd rather not worry about it.

rst_files/kalman.rst
arnavs: Done.

rst_files/kalman.rst
Nosferican: The levels are off. In the Jupyter notebook, every line except the first has a leading two spaces.

rst_files/kalman.rst
arnavs: This one's a bit tough because it's a container of arrays, so `zeros` doesn't apply. 

I changed it to:

```
temp = [ [1.0], [2.0], [3.0], [4.0], [5.0] ]
densities = similar(temp); # we'll do 5 rounds of updating
```

rst_files/kalman.rst
arnavs: Because this is a sequential computation (as opposed to strictly iterating over some static object, which can be done asynchronously), I think the logic as clearer as a loop.

arnavs:
Done @Nosferican. Plots we can fix later (@XiaojunGuan), same with alignment (and per style guide, don't think it's strictly necessary). Agree that we should eventually replace these images with Julia code.

arnavs:
@Nosferican I thought those are OK, because we're testing set membership?

jlperla:
I think those are effectivly treating the zip like an iterator.  There is no `set` like that in any wequations, for example.

jlperla:
Another thing.  Can we avoid the `abs2` function whenever possible?

arnavs:
Done and done.

jlperla:
i.e. squared is much easier to write, and 
```
julia> @btime abs(1.0)
  0.001 ns (0 allocations: 0 bytes)
1.0

julia> @btime 1.0^2
  0.001 ns (0 allocations: 0 bytes)
1.0

julia> @btime (-1.0)^2
  0.001 ns (0 allocations: 0 bytes)
1.0
```

rst_files/kalman.rst
```diff
@@ -738,8 +738,8 @@ Exercise 3
         # update state and record error
         Ax = A * x
         x = rand(MultivariateNormal(Ax, Q))
-        e1[t] = sum(abs2(a - b) for (a, b) in zip(x, kn.cur_x_hat))
-        e2[t] = sum(abs2(a - b) for (a, b) in zip(x, Ax))
+        e1[t] = sum(abs(a - b)^2 for (a, b) in zip(x, kn.cur_x_hat))
+        e2[t] = sum(abs(a - b)^2 for (a, b) in zip(x, Ax))
```
Nosferican: No need for `abs` if you going to `x^2`. `(a-b)^2`

arnavs: That's true as well.

jlperla:
I think there are places in libraries and super-efficient kernels where using `abs2` makes sense, but otherwise let me put in a new issue to replace it.

Nosferican:
Refer to the style for comments in the code (only when vital and favor explanation in the markdown cells).

arnavs:
Thanks for the style fixes @Nosferican. I'm going to merge this now since I think it's in a good enough state to generate HTML. Will let an RA know to enforce the comments rule, etc.

PR #210, opened 2018-10-25
Re-Write LLN-CLT

Nosferican:


rst_files/lln_clt.rst
```diff
@@ -217,82 +202,58 @@ The dots represent the underlying observations :math:`X_i` for :math:`i = 1, \ld
...
+    gr(fmt = :png, size = (900, 500))
 
-        dist_data[i, :] = data'
-        push!(sample_means, sample_mean)
-        push!(dist_means, m * ones(n))
-        push!(titles, name)
+.. code-block:: julia
 
+    function ksl(distribution, n = 100)
+        title = replace(replace(string(distribution), r"\{.*\}" => ""), r".=" => "")
```
arnavs: This is a bit clever, could we try something with `distribution.type` which returns a `Core.TypeName`. Maybe this can be stringified?

arnavs:
Thanks @Nosferican, I think this is both accessible and a big improvement. If it's too much trouble to get rid of the title thing above, we can just merge.

PR #218, opened 2018-10-26
Rewrite "Job Search I: The McCall Search Model"

arnavs:
Has all of our big changes (NLsolve, Expectations, Structs, etc.) @jlperla I think the rewrite you sent me will come into play for McCall 2.

### DIFF
--- a/rst_files/mccall_model.rst
+++ b/rst_files/mccall_model.rst
@@ -311,7 +311,8 @@ Let's start with some imports
 
 .. code-block:: julia
 
-    using Plots, QuantEcon, Distributions, LaTeXStrings, LinearAlgebra, Random
+    using StatPlots, Plots, QuantEcon, Distributions, LinearAlgebra, Random, Expectations, NLsolve
+
     gr(fmt=:png)
 
 Here's the distribution of wage offers we'll work with
@@ -320,20 +321,15 @@ Here's the distribution of wage offers we'll work with
 .. code-block:: julia
 
     n, a, b = 50, 200, 100
-    w_min, w_max = 10, 60
-    const w_vals = range(w_min,  w_max, length = n+1)
     dist = BetaBinomial(n, a, b)
-    const p_vals = pdf.(dist, support(dist))
 
-    plt = plot(w_vals, p_vals, label = L"$p_i$", xlabel = "wages", ylabel = "probabilities")
+    plt = plot(dist, xlabel = "wages", ylabel = "probabilities", legend = false)
 
 
 .. code-block:: julia
     :class: test
 
-    @testset "First Plot Test" begin
-        @test p_vals[4] ≈ 5.789516415478556e-17 # p invariance
-        @test w_vals[1] == 10 && w_vals[end] == 60 && length(w_vals) == 51 # grid invariance
+    @testset begin
         @test dist isa Distributions.BetaBinomial && dist.n == 50 && dist.α == 200 && dist.β == 100 # distribution invariance
     end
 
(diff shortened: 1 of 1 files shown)


Nosferican:
Add a test to the final cell based on `stop_times`

arnavs:
Thanks @Nosferican, added that test cell.

rst_files/mccall_model.rst
```diff
@@ -347,64 +319,36 @@ Our initial guess :math:`v` is the value of accepting at every given wage
...
+    c = 25
+    β = 0.99
+    num_plots = 6
+    E = expectation(dist)
+
+    # functions and matrix to plot
+    stop_val(w) = w / (1 - β)
+    cont_val(v) = c + β * E * v # using E as a linear operator
+    best_val(w, v) = max(stop_val(w), cont_val(v))
+    vs = repeat(w_vals ./ (1 - β), outer = (1, 6)) # data to fill
```
jlperla: I don't find this very clear.  What does the `outer = (1, 6)` do here?  Why are we willing in a bunch of information for `vs[:, col]` for `col > 1` we are about to replace?  If this is just a question of preallocating a matrix, we could make that more clear.

rst_files/mccall_model.rst
```diff
@@ -347,64 +319,36 @@ Our initial guess :math:`v` is the value of accepting at every given wage
...
+    # parameters and constant objects
+    w_min, w_max = 10, 60
+    w_vals = range(w_min, w_max, length = n + 1)
+    c = 25
+    β = 0.99
+    num_plots = 6
+    E = expectation(dist)
+
+    # functions and matrix to plot
+    stop_val(w) = w / (1 - β)
```
jlperla: Having the `stop_val` and `cont_val` just seems to make the code further from the math.  Why not combine into the `best_val` function and get rid of them, or am I missing something?

Finally, instead of `best_val` why not make it closer to the math by calling it `T`?  That is what it does?

rst_files/mccall_model.rst
```diff
@@ -347,64 +319,36 @@ Our initial guess :math:`v` is the value of accepting at every given wage
...
-
+    plot(vs)
 
 Here's more serious iteration effort, that continues until measured deviation
 between successive iterates is below `tol`
 
-
 .. code-block:: julia
 
-    function compute_reservation_wage(c, β;
```
jlperla: I think it would be useful to resurrect this for teaching purposes.  Note that "The Algorithm" section, and (3) in particular, has this built in as an iteration.  We can then show the two different ways of doing it, where we show a particular, naive, fixed point algorithm and then call the library.  We don't need to do this for every time, but I think it would be helpful here.

rst_files/mccall_model.rst
```diff
@@ -602,30 +515,23 @@ Repeat a large number of times and take the average
...
-
 .. code:: julia
 
     function compute_stopping_time(w_bar; seed=1234)
-
         Random.seed!(seed)
         stopping_time = 0
-
         t = 1
         while true
```
jlperla: Is there a hidden error here if they change the parameters and have parameters such that w_bar is never binding?  Might want to make it more robust.

rst_files/mccall_model.rst
arnavs: You're right. Will implement these changes.

rst_files/mccall_model.rst
arnavs: The `outer = (1, 6)` is the Julian syntax for `repeat` now (I'm not a fan of it either). 

I'll make that clearer. We could fill it with `0`s, but the first column needs to be pre-allocated with the first guess, and I figured it was easier to just fill the whole thing.

Nosferican: For an `Array`, you can use `repeat(w_vals ./ (1 - β), 1, 6)` # Repeat 1 in the first dimension and 6 times on the second.

jlperla: Sorry, I meant to "What does the `outer = (1, 6)` do?" with the tone of how a reader might express confusion.  I understand (i.e. had to lookup :-) the notation.

My main issue was that we fill in a big array but everything is overwritten except the first row.  The use of `repeat` here masks that, and makes you think that the data in the other rows actually matters.  Better to preallocate the matrix and directly fill in the first row to make that clearer.

rst_files/mccall_model.rst
arnavs: You're right. I'll include a check that the `w_bar` is inside the distribution.

rst_files/mccall_model.rst
arnavs: Added this (currently in a test block), pending what you want to do with it.

Issue #189, opened 2018-10-22
High degradation of plot quality with GR


Nosferican:
Some plots that have been ported from other backends to GR make the new plots utterly useless. Some can be "fixed" through a re-write while others simply need a different approach given the limitations of the backend. Opening this issue to keep track of the decisions.

jlperla:
Can you give an example of this?  Some of the problems are because Plots.jl had some bugs with sizes of labels/etc. that are being fixed.

Also I think we need to make a decision on whether we go with `png` or `svg`.... At this point
`gr(fmt=:png)` seems to solve a lot of issues with the sizes, and it also makes the notebooks show pictures without trusting it first.  A major benefit for users.

Are there any downsides?

Nosferican:
![clt3d](https://user-images.githubusercontent.com/2529329/47309533-4cb0de00-d5fa-11e8-91ec-045175b4c6d0.png)
![gr](https://user-images.githubusercontent.com/2529329/47309535-4f133800-d5fa-11e8-9ca1-6f07f64b996e.png)

jlperla:
Both are with GR, and the second one is the version with `png`?  Are all of the major issues you are seeing with 3D plots

arnavs:
If I recall, we talked about these issues before and they all bound to 3D plots only.

jlperla:
There aern't all that many 3D plots, and some of them could even be shown as a sequence of 1D plots if required.

Why not identify all of the terrible plots for now in a checklist here, and then we can one-by-one see if  we can make it work with GR in its current form.

BTW, I had a student today with terrible (2D) plots, but when he used the `gr(fmt = :png)` it started working again.  Should we make a global change?  If nothing else, it means that when we get sent notebooks from students we can actually see the graphs without `Trust`

arnavs:
@jlperla I think we already made the global change in #171.

jlperla:
Great.  Even if `png` is slightly inferior to `svg`, it is nice to have consistency for testing purposes.  But yeah, why don't we keep a list of ugly figures to fix up there.  This is high priority, especially if the HTML rendering comes soon'ish.

Nosferican:
The first one is the 0.6 from the webpage. The second one is with `gr(fmt=:png)`.
This is another example,
From webpage
![lapro](https://user-images.githubusercontent.com/2529329/47312828-0613b180-d603-11e8-9ebc-f36af71f23ab.png)
After porting to GR
[...]

PR #220, opened 2018-10-27
Schelling plot display fixed

ajozefiak:
The display function was not displaying the plots when running this lecture. I modified the lecture to collect the generated plots into an array and plotting it afterwards. 

This change is motivated by #189


arnavs:
Thanks, and nice use of splatting.

PR #222, opened 2018-10-29
Re-write Finite Markov

Nosferican:
Using dictionaries rather than temporary text files to disk.

### DIFF
--- a/rst_files/finite_markov.rst
+++ b/rst_files/finite_markov.rst
@@ -1288,8 +1288,12 @@ Exercise 2
 .. code-block:: julia
 
     nodes = keys(web_graph_data)
+    n = length(nodes)
     # create adjacency matrix of links (Q[i, j] = true for link, false otherwise)
-    Q = mapreduce(edges -> (nodes .∈ Ref(edges))', vcat, values(web_graph_data))
+    Q = fill(false, n, n)
+    for (node, edges) in enumerate(values(web_graph_data))
+        Q[node, nodes .∈ Ref(edges)] .= true
+    end
 
     # create the corresponding stochastic matrix
     P = Q ./ sum(Q, dims = 2)
--- a/rst_files/finite_markov.rst
+++ b/rst_files/finite_markov.rst
@@ -36,10 +36,9 @@ Setup
 
 .. code-block:: julia
 
-    # Import packages.
-    using LinearAlgebra, Statistics, Compat
-    using QuantEcon, Random, Plots, Printf, Distributions
-    gr(fmt=:png) # Setup for the Plots.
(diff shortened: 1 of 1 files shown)


rst_files/finite_markov.rst
```diff
@@ -1294,79 +1268,35 @@ compare it to the stationary probability.
...
-    edges = readlines(infile)
-    for edge in edges
-        from_node, to_node = [String(m.match) for m = eachmatch(r"\w", edge)] # String() required to go from Substring to String. r"\w" is a word character.
-        i = findfirst(isequal(from_node[1]), alphabet) # [1] required to go from String to Char
-        j = findfirst(isequal(to_node[1]), alphabet)
-        Q[i, j] = 1
-    end
+    nodes = keys(web_graph_data)
+    # create adjacency matrix of links (Q[i, j] = true for link, false otherwise)
+    Q = mapreduce(edges -> (nodes .∈ Ref(edges))', vcat, values(web_graph_data))
```
jlperla: Cool.  But *WAY* too tricky for introductory programmers.  I think the basic test is: if `reduce` and especially `mapreduce` and `vcat` or `hcat` are used, then there is no chance introductory users will understand it.  I think we just need to entirely hold off on functional programming patterns of this sort.

BTW, I think that in a future iteration of the lectures, we should add in a "functional programming and parallel computing" section where we can go through these sorts of programming patterns carefully, and explain why they are so essential to get parallel computing working well.

I really like the separation of the file loading from the creation of the graph is a major improvement, though.

jlperla: But I think it is worth leaving in that line as code, commented out, to students with a comment that says something to the effort of "functional programming style can be very concise, if you invest some time in learning it"

jstac: It's neat code and some people love functional programming -- although I agree with @jlperla  that it  simple explicit loops will be better for the basic implementation.

Perhaps what's needed is a short lecture on functional programming in Julia that's flagged at the top as more advanced.  This would be a good example to use in that lecture.

jlperla: Also, in this specific case, my suspicion is that - even conditioning on following a lecture on functional programming, a more direct loop and iterator approach is better for the notes.

I generally like functional techniques for my own stuff, but fully understanding that code and convincing myself it is correct would take me minimum 5+ minutes of playing with MWE in the REPL.. And I would probably end up writing the loop version out of paranoia in a unit test.  So, even if I needed it to be functional to parallelize, I would probably have written the loop version as a check 

This is probably an example of where brevity does not bring clarity.

arnavs:
Merging now since @Nosferican addressed feedback.

PR #224, opened 2018-10-29
Rewrite McCall II

arnavs:
Fixedpoint, expectations, etc. 

Will incorporate comments from @jlperla and then this should be good to review.

### DIFF
--- a/rst_files/mccall_model_with_separation.rst
+++ b/rst_files/mccall_model_with_separation.rst
@@ -166,20 +166,6 @@ and
       \beta (1 - \gamma) U
           + \beta \gamma \sum_i \max \left\{ U, V(w_i) \right\} p_i
 
-Formally, we can define a "Bellman operator" T which maps:
-
-.. math::
-    :label: bellman_operator
-          
-    TV(\cdot) = u(\cdot) + \beta  (1-\alpha)V(\cdot) + \alpha U 
-          
-In which case we are searching for a fixed point
-    
-.. math::
-    :label: bellman_operator_fp
-          
-    TV^{*} = V^*              
-
 Solving the Bellman Equations
 -------------------------------
 
@@ -216,6 +202,21 @@ and
 
 
 starting from some initial conditions :math:`U_0, V_0`
+
+Formally, we can define a "Bellman operator" T which maps:
+
+.. math::
+    :label: bellman_operator
+          
+    TV(\cdot) = u(\cdot) + \beta  (1-\alpha)V(\cdot) + \alpha U 
+          
+In which case we are searching for a fixed point
+    
+.. math::
+    :label: bellman_operator_fp
+          
+    TV^{*} = V^*              
+
 
 As before, the system always converges to the true solutions---in this case,
 the :math:`V` and :math:`U` that solve :eq:`bell01_mccall` and :eq:`bell02_mccall`
--- a/rst_files/mccall_model_with_separation.rst
+++ b/rst_files/mccall_model_with_separation.rst
@@ -168,18 +168,17 @@ and
 
 Formally, we can define a "Bellman operator" T which maps:
 
-    .. math::
-        :label: bellman_operator
+.. math::
+    :label: bellman_operator
           
-        TV(\cdot) = u(\cdot) + \beta  (1-\alpha)V(\cdot) + \alpha U ]
+    TV(\cdot) = u(\cdot) + \beta  (1-\alpha)V(\cdot) + \alpha U 
           
 In which case we are searching for a fixed point
     
-    .. math::
-        :label: bellman_operator
+.. math::
+    :label: bellman_operator_fp
           
-        TV^{*} = V^*
-              
+    TV^{*} = V^*              
 
 Solving the Bellman Equations
 -------------------------------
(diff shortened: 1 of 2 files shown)


arnavs:
@jlperla Resolved the issues in McCall I. I think this one is ready for review now? (Do we also need to resurrect some iteration here?)

jlperla:
I don't think so. Worth doing it once for expostion, and to help reinforce making algorithms orthogonal in generic programming, but otherwise not sure it is worth it.

But submit that PR for NLsolve default parameters so that people get the (correct) impression that the function "just works" and is faster without tinkering

Nosferican:
Check the change log summary of [3add3e8]. It has a few issues that should help with this and future re-writes.

rst_files/mccall_model_with_separation.rst
```diff
@@ -447,17 +434,17 @@ we can create an array for reservation wages for different values of :math:`c`,
 
     for (i, c) in enumerate(c_vals)
         mcm = McCallModel(c = c)
-        w_bar = compute_reservation_wage(mcm)
+        ~, ~, w_bar = solve_mccall_model(mcm)
```
Nosferican: ?

rst_files/mccall_model_with_separation.rst
```diff
@@ -216,55 +216,51 @@ Implementation
...
+        # parameter validation
+        @assert c > 0.0
+        @assert minimum(w) > 0.0 # perhaps not strictly necessary, but useful here 
+        
+        # necessary objects 
+        u_w = u.(w, σ) 
+        u_c = u(c, σ)
+        E = expectation(dist) # expectation operator for wage distribution
+
+        # Bellman operator T. Fixed point is x* s.t. T(x*) = x*
```
Nosferican: Why not use a more descriptive name than `T` unless used as such in the math?

rst_files/mccall_model_with_separation.rst
```diff
@@ -274,6 +270,19 @@ Let's plot the approximate solutions :math:`U` and :math:`V` to see what they lo
...
+
+    # a default utility function
+    u(c, σ) = (c^(1 - σ) - 1) / (1 - σ) 
+
+    # default objects
+    const n = 60                                           # n possible outcomes for wage
+    const default_w_vec = range(10, 20, length = n)        # wages between 10 and 20
+    const a, b = 600, 400                                  # shape parameters
+    const dist = BetaBinomial(n-1, a, b)                   # wage distribution
+
```
Nosferican: The these globals aren't really needed at all. Best to drop these.

rst_files/mccall_model_with_separation.rst
```diff
@@ -274,6 +270,19 @@ Let's plot the approximate solutions :math:`U` and :math:`V` to see what they lo
...
+    # model constructor
+    McCallModel = @with_kw (α = 0.2, β = 0.98, γ = 0.7, c = 6.0, σ = 2.0, u = u, w = default_w_vec, dist = dist)
```
Nosferican: Maybe line break these and put comments on each.

rst_files/mccall_model_with_separation.rst
Nosferican: Julian way would be to use keywords to determine optional outputs.

rst_files/mccall_model_with_separation.rst
jlperla: Yeah, the style guide had no `const` as well

rst_files/mccall_model_with_separation.rst
jlperla: Hopefully the math had T or something similar. If not, we could add the operator in to make it clear.. Usually it was called T in other lectures?

rst_files/mccall_model_with_separation.rst
arnavs: We rewrote this function to put everything under one roof (inside a closure), and to return all key objects (`V, U, w_bar`). I think that this is more clear than muddying up the code with `if/else` flags...

rst_files/mccall_model_with_separation.rst
arnavs: I think we should add the operator to the math on this one. But @Nosferican, I also think the `x = Tx` convention for the iteration is commonplace enough that it should be clear what this is doing.

rst_files/mccall_model_with_separation.rst
arnavs: Done.

rst_files/mccall_model_with_separation.rst
arnavs: Alternately, we could return a NT `(V = V, U = U, w_bar = w_bar)`, but I think this is OK for now.

Nosferican: That just smells like nargout *cough* MATLAB *cough* If not mistaken there is only one instance besides the default, keyword arguments are not too bad.

Nosferican:
Let me know when this is ready for review. Do take a look at the changes I did in the older branch and let me know.

rst_files/mccall_model_with_separation.rst
jlperla: Having optional outputs is just not important here. I think this is only written this way because of how the discussion around the code was structured. I may change that organization

arnavs:
@Nosferican I think most of the changes were obsoleted by what Jesse asked for, but I took another look and made some edits (like using `fill`, etc.)

rst_files/mccall_model_with_separation.rst
jlperla: Also, I love named tuple returns when there lots of things returned

jlperla:
Sorry, I have to merge this because otherwise I am unable to concurrently generate notebooks for teaching.  I will make modifications to master, and you can do the same (or put other tweaks in a PR to discuss).

rst_files/mccall_model_with_separation.rst
arnavs: I made this change to the lecture, and will add this to the style guide.

PR #230, opened 2018-10-31
career.rst PyPlot to Plots (with some stylistic issues)

ajozefiak:
I've converted the PyPlot API to Plots, however, the surface plot has a few issues like mapping the y-label to the z-axis and having the surface rotated in a poor angle. The contour plot is fine, though could be made better.


arnavs:
No problem, we can tweak that. Thanks so much.

PR #234, opened 2018-10-31
Cosmetics and Bugfix

arnavs:
Wanted to confirm that this approach to resolving the bug (that the defined `T` was capturing globals upon definition) was satisfactory.

### DIFF
--- a/rst_files/mccall_model.rst
+++ b/rst_files/mccall_model.rst
@@ -310,30 +310,49 @@ We can explore taking expectations over this distribution
     # exploring the properties of the operator
     wage(i) = w[i+1] # +1 to map from support of 0
     E_w = E(wage)
-    E_w_2 = E(i -> (w[i+1])^2 - E_w^2) # variance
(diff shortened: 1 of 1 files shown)


rst_files/mccall_model.rst
```diff
@@ -310,30 +310,49 @@ We can explore taking expectations over this distribution
...
-To impllement our algorithm, let's have a look at the sequence of approximate value functions that
+To implement our algorithm, let's have a look at the sequence of approximate value functions that
 this fixed point algorithm generates
 
 Default parameter values are embedded in the function
 
 Our initial guess :math:`v` is the value of accepting at every given wage
 
-.. code:: julia
+.. code-block:: julia
```
arnavs: Changed this for consistency with the rest of the lectures @jlperla. I don't think it made any discernible difference, but if you'd prefer I can change it back.

jlperla:
I think I would prefer that you dropped this whole PR and made changes to my final version.  I am modifying too many of these things.

PR #241, opened 2018-11-02
Parameters Pass for "Optimal Savings III: Occasionally Binding Constraints"

arnavs:
Not much going on in this one; there was one struct that wasn't mutated. 

#135


Nosferican:
See if everything is good. Made a note of the missing maximize pass.

rst_files/ifp.rst
```diff
@@ -48,18 +48,11 @@ This method turns out to be
 
 * More efficient numerically than value function iteration
 
-
 References
 ----------------
 
 Other useful references include :cite:`Deaton1991`, :cite:`DenHaan2010`, :cite:`Kuhn2013`, :cite:`Rabault2002`,  :cite:`Reiter2009`  and :cite:`SchechtmanEscudero1977`
```
arnavs: @Nosferican I think the convention is to put the setup right at the end of the overview section, as opposed to before the first code block. But I don't have strong feelings about it.

Nosferican: It does make working the lectures a pain. In the Setup discussion I had indicated to have them right before the `using Test` (right before any code execution). The Setup is only relevant to executing code, why should it be positioned between a bunch of markdown cells and actual code? It also forces to go up to the top to run that one cell before being able to work with the code.

arnavs: Yeah, I'm fine with either. Let's see what @jlperla thinks.

Nosferican:
I am arguing that the Setup include should be right before the `using Test` cell (right before the first code cell) instead of currently where it stands before the markdown cells explaining the lesson and the start of the code cells.

arnavs:
@Nosferican checked with Jesse, and I don't think he minds too much. So this should be fine. Will merge now.

PR #242, opened 2018-11-02
Parameters Pass for "Asset Pricing II: The Lucas Asset Pricing Model"

arnavs:
Likewise as in #241. 

Partially resolves issue #135.


arnavs:
Thanks, looks great.

PR #247, opened 2018-11-03
RW Schelling Parameters Pass

Nosferican:


rst_files/schelling.rst
```diff
@@ -143,76 +128,62 @@ Here's one solution that does the job we want. If you feel like a
...
-        end
 
-        # == Sort from smallest to largest, according to distance == #
-        sort!(distances)
+        distances = get_distance.(Ref(a), others)
 
-        # == Extract the neighboring agents == #
-        neighbors = [agent for (d, agent) in distances[1:num_neighbors]]
+        # Extract the neighboring agents
+        neighbors = agents[partialsortperm(distances, 1:neighborhood_size)]
```
arnavs: I think the old way was clearer.

rst_files/schelling.rst
```diff
@@ -143,76 +128,62 @@ Here's one solution that does the job we want. If you feel like a
...
-        nothing
-    end
+    draw_location!(a) = a.location .= rand(2)
 
     # distance is just 2 norm: uses our subtraction function
     get_distance(a, o) = norm(a.location - o.location)
 
-    function is_happy(a, others)
-        "True if sufficient number of nearest neighbors are of the same type."
+    function is_happy(a, others, neighborhood_size, preference)
```
arnavs: I think this is best written as it was before; i.e., `is_happy(agent, others)`

rst_files/schelling.rst
```diff
@@ -143,76 +128,62 @@ Here's one solution that does the job we want. If you feel like a
...
     end
 
     function plot_distribution(agents, cycle_num)
-        x_vals_0, y_vals_0 = Float64[], Float64[]
-        x_vals_1, y_vals_1 = Float64[], Float64[]
+        x_vals_0, y_vals_0 = zeros(0), zeros(0)
+        x_vals_1, y_vals_1 = zeros(0), zeros(0)
 
-        # == Obtain locations of each type == #
+        # Obtain locations of each type
```
arnavs: Comments should be lowercase

rst_files/schelling.rst
```diff
@@ -266,10 +230,12 @@ then increase the number of agents.
         end
     end
 
-    println("Converged, terminating")
-    plot(plot_array..., layout=(count-1,1), size=(600,(count-1)*400))
+    plot(plot_array...,
+         layout = (length(plot_array), 1),
+         size = (600, 400 * length(plot_array)),
+         title = permutedims(string.("Cycle ", 1:length(plot_array))))
```
jlperla: If we are going to use permutedims, I think we need to move it to a separate line before the `plot` and have a comment explaining what it is doing.  Is it really worth it, though, or is there an easier thing to explain that may have an additional line of code or two?

arnavs: I think that we should hold off on using complicated code just for plots. We had a lot of `reshape`s and splatting in here as well.

A list comprehension probably works well for generating plot titles procedurally, e.g. `titles = ["cycle $n" for n in 1:num_plots]` or whatever

Nosferican: `permutedims` is the with the `(2,1)` is same as `transpose` or `'`, but it works on nun-numerics (non-recursive). I think a one line comment should be able to communicate this.

jlperla: Yeah.  I think the general rule should be: if there is the word `permute` or `perm` in the julia function name, then we should ask whether it really is necessary.  Comprehensions or an extra line of code can make things easier to read.

rst_files/schelling.rst
Nosferican: I think this should be a style guide, but using enclosure functions is kind of risky and should be avoided if possible. Having all the arguments to fully characterize the function also helps for transferring parts of code.

rst_files/schelling.rst
jlperla:
```
["cycle $n" for n in 1:length(plot_array)]
```
vs.
```
 permutedims(string.("Cycle ", 1:length(plot_array)))
```
... even with a comment, the former is easier to read.  In-built sorting functions, permutations, etc. are tricky and would require people to lookup how they work.

Nosferican: The one above would not work. It would be something like
```
reduce(hcat, "cycle $n" for n in eachindex(plot_array))
```
which I would be fine with.

rst_files/schelling.rst
jlperla: I think pretty much the opposite.  We have been making heavy use of closures, and should continue to do so.  The key is that you can copy the whole code inside of a function to transfer it.

If we are fixing the `neighborhood_size` and `preference` and just applying it to `a` and `others` then it is a lot more direct to use a closure.

It is also easier to read, since the reader can see what is really treated as a function argument.

rst_files/schelling.rst
jlperla: `reduce` is discouraged and `reduce(hcat` should be avoided at all costs.  I would rather have 2-3 lines of code that are clear than trying to cram everything into a fancy function.

Nosferican: Is `reshape` fine?
```
reshape(["cycle $n" for n in eachindex(plot_array)], 1, 5)
```

arnavs: I don't think it's useful; I would take it out and nothing changed.

Nosferican: Eliminate the titles?

jlperla: I think your `reshape` is a good approach.  It is easy to know what the function does just by the title.

arnavs: @Nosferican Just check to make sure the `reshape` is necessary. I would take it out in some of the other lectures and it didn't break anything. 

Will go offline to work on another project for a bit, but I guess my big feedback here is that I think this is a lecture where the `mutable struct` isn't too bad of an idea.

arnavs:
Thanks @Nosferican for addressing feedback. I trimmed a few comments from the `is_happy` function, because it was looking a bit intimidating. Merging now.

arnavs:
Note that, as written, this function doesn't have things inside a closure (which I think is the best way to do this), but rather relies on a global `agents` vector. For v1.1, I think this should be changed per the above.

jlperla:
@arnavs  Unless I am missing something, though, this isn't a problem.

The key is not that the code currently is a closure, but rather that it is structured so that it *could* be a closure.  i.e. the test would be to copy the code into a function and execute it, then it would work.  I could be wrong, but I believe that the way @Nosferican  wrote it would fulfill that requirement.

PR #248, opened 2018-11-03
RW Markov Asset (Parameters)

Nosferican:


rst_files/markov_asset.rst
```diff
@@ -589,57 +543,40 @@ the type `AssetPriceModel`
...
 
-
-    # price/dividend ratio of the Lucas tree 
-    function tree_price(ap::AssetPriceModel)
-        # == Simplify names, set up matrices  == #
-        β, γ, P, y = ap.β, ap.γ, ap.mc.p, ap.mc.state_values
+    # price/dividend ratio of the Lucas tree
+    function tree_price(ap; γ = ap.γ)
+        # Simplify names, set up matrices
+        β, P, y = ap.β, ap.mc.p, ap.mc.state_values
```
arnavs: This should be an `@unpack`, i.e. 

```
@unpack β, P, state_values = ap 
```

rst_files/markov_asset.rst
```diff
@@ -847,42 +765,38 @@ into vector :math:`Tw` via
...
 
-
 .. code-block:: julia
 
     # price of perpetual call on consol bond
-    function call_option(ap::AssetPriceModel, ζ::AbstractFloat, p_s::AbstractFloat, ϵ=1e-7)
+    function call_option(ap, ζ, p_s, ϵ = 1e-7)
 
-        # == Simplify names, set up matrices  == #
+        # Simplify names, set up matrices
```
arnavs: Comments should be lowercase

rst_files/markov_asset.rst
Nosferican: That struct uses `ap.mc.val`, not `ap.val` so `@unpack` doesn't work in this case.

rst_files/markov_asset.rst
Nosferican: Best to do that in another pass. I think most if not all comments at least in this lesson are camel case.

rst_files/markov_asset.rst
jlperla: I think we should unpack to beta and mc, and then can do the mc stuff in a separate line. I think it would be more clear and consistent with your style

arnavs:
Thanks @Nosferican, looks good

PR #252, opened 2018-11-03
RW Uncertainty_Traps (Parameters)

Nosferican:
- This one keeps the `mutable struct` since the code goes pretty well with an `update!` function.
- Please check the before last plot as from the transition from Gadfly it looks different.

### DIFF
--- a/rst_files/uncertainty_traps.rst
+++ b/rst_files/uncertainty_traps.rst
@@ -2,15 +2,14 @@
 
 .. include:: /_static/includes/lecture_howto_jl.raw
 
-***************************************
+*****************
 Uncertainty Traps
-***************************************
+*****************
 
 .. highlight:: julia
 
-
 Overview
-============
+========
 
 In this lecture we study a simplified version of an uncertainty traps model of Fajgelbaum, Schaal and Taschereau-Dumouchel :cite:`fun`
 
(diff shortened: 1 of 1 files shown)


jlperla:
I took at look at the code.   I don't think we should be using mutable structs here, and reorganizing it is worth it.

jlperla:
Another thing on this: I think we should be willing to use dataframes more often than we have. Pushing things onto dataframes, as we calculate, might help things.

One of the great things about dataframes is you can push a named tuple onto them. It is pretty slick.

jlperla:
Except that the whole lecture is basically about doing a simulation of a process and then plotting the results from a dataframe.

If that is what we are doing, then pushing things onto the dataframe as we do a calculation isn't the worst idea.  Regardless, I really don't like mutable structs in this case.

arnavs:
@Nosferican Do you have the bandwidth right now to work on this, or should I take a pass

Nosferican:
Go for it. I still think this is a nice case for mutable structs, but curious to see how you come up with a different approach.

jlperla:
This can wait. It works, and there are plenty of other things to work on

arnavs:
Sounds good @jlperla. Then I suggest merging this (because it contains a lot of other improvements) and revisiting the `mutable struct` later. Also helps cut down on the number of active branches :)

PR #254, opened 2018-11-03
RW Aiyagari (Parameters)

Nosferican:
- Needs a fix of the link in the second cell

### DIFF
--- a/rst_files/aiyagari.rst
+++ b/rst_files/aiyagari.rst
@@ -4,14 +4,12 @@
 
 .. highlight:: julia
 
-
-*******************************************************
+******************
 The Aiyagari Model
-*******************************************************
-
+******************
 
 Overview
-============
+========
 
 In this lecture we describe the structure of a class of models that build on work by Truman Bewley :cite:`Bewley1977`
 
@@ -31,7 +29,6 @@ The model features
 
 * Limits on amounts individual agents may borrow
 
-
 The Aiyagari model has been used to investigate many topics, including
 
 * precautionary savings and the effect of liquidity constraints :cite:`Aiyagari1994`
@@ -43,7 +40,7 @@ The Aiyagari model has been used to investigate many topics, including
 * etc., etc., etc.
 
 References
--------------
+----------
 
 The primary reference for this lecture is :cite:`Aiyagari1994`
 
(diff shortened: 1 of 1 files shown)


rst_files/aiyagari.rst
```diff
@@ -4,14 +4,12 @@
 
 .. highlight:: julia
 
-
-*******************************************************
+******************
```
arnavs: @Nosferican the reason I make these long are because Sphinx actually cares how long they are (i.e., if the text is longer than the `**********`, it'll throw an error for the title not being long enough).

Nosferican: Exactly, just make them as long as needed

Good
```
This is a title
===============
```
Bad
```
This is a title
========================================
```
Best to be explicit than ambiguous.

arnavs: OK, put it in the style guide I guess

Nosferican: We should have an internal style guide for the RST (the current style if for the Jupyter output)

rst_files/aiyagari.rst
```diff
@@ -227,64 +198,46 @@ Here we take the state to be :math:`s_t := (a_t, z_t)`, where :math:`a_t` is ass
...
+                          a_min = 1e-10,
+                          a_max = 18.0,
+                          a_size = 200,
+                          a_vals = range(a_min, a_max, length = a_size),
+                          z_size = length(z_chain.state_values),
+                          n = a_size * z_size,
+                          s_vals = gridmake(a_vals, z_chain.state_values),
+                          s_i_vals = gridmake(1:a_size, 1:z_size),
+                          u = σ == 1 ? x -> log(x) : x -> (x^(1 - σ) - 1) / (1 - σ),
+                          R = setup_R!(fill(-Inf, n, a_size), a_vals, s_vals, r, w, u),
```
arnavs: Can the `-Inf` here be replaced with `0`?

Nosferican: The `-Inf` is suppose to represent a really really bad utility (no consumption, then the agent dies). It follows the math. Zero doesn't carry that connotation.

rst_files/aiyagari.rst
arnavs: I think you can just add a section to the existing document. No need to make a whole new one (i.e., if you're contributing rST, make sure you...)

rst_files/aiyagari.rst
arnavs: Gotcha, thanks. I'll throw in a comment about it.

PR #267, opened 2018-11-08
[CONF] add s3 urlprefix 

mmcky:
This PR adjusts `conf.py` to add a url reference for static assets in generated notebooks. 

Current location is `s3` compare-lectures bucket. 

This will need to be updated in the future. 

This will require the latest `sphinxcontrib-jupyter` version.

### DIFF
--- a/rst_files/conf.py
+++ b/rst_files/conf.py
@@ -421,3 +421,6 @@ jupyter_target_html = True
 
 #Drop Tests Embedded in Lectures
 jupyter_drop_tests = False
+
+#Use urlprefix images
+jupyter_images_urlpath = "https://s3-ap-southeast-2.amazonaws.com/compare-lectures.quantecon.org/jl/_static/"


jlperla:
If this is merged, does it mean we will be able to use the notebooks with links not in _static?  Might be work checking sooner rather than later.

mmcky:
hey @arnavs no update should be required as the `prefix` should take you to the relative root level (as per the relative link).

mmcky:
@jlperla what links (not in ``_static``) do you have in mind? Can you give me an example?

jlperla:
I am not sure what I meant... but I think I meant testing?  Basically, take the notebook output which has images refereed, move a file so that it no longer has any local folders accessible, and see if it works.  RIght now we need to distribute the whole `_static` folder with the notebook to see images.

mmcky:
if we use the prefix then the whole contents of `_static` can be referred via the internet with this setup. That reminds me though we probably need to add prefix options for:
- [ ] `_downloads` for internal links to files.

arnavs:
Thanks @mmcky, I think this looks good and the output works, so we should merge it. Will create a separate issue for these checkmarks/TODOs.

PR #282, opened 2018-11-10
RW ODU

Nosferican:
- Needs Parameters
- Interpolation update

### DIFF
--- a/rst_files/odu.rst
+++ b/rst_files/odu.rst
@@ -5,15 +5,14 @@
 
 .. highlight:: julia
 
-******************************************
+************************************
 Job Search III: Search with Learning
-******************************************
+************************************
 
 .. contents:: :depth: 2
 
 Overview
-============
-
+========
 
 In this lecture we consider an extension of the :doc:`previously studied <mccall_model>` job search model of McCall :cite:`McCall1970`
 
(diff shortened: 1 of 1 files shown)


arnavs:
@Nosferican I did the interpolations stuff on `master` in https://github.com/QuantEcon/lecture-source-jl/commit/943d0c3984223ee83470a1fc3231f37a5d114d5d. So this may need to be rebased

arnavs:
Hey @Nosferican I'm going to close this, since I think the one in `master` is more up-to-date and incorporates all of the non-aesthetic changes? But let me know if you disagree.

Nosferican:
Are the ones in master with the correct alignment. We should make a pass to ensure all indentation is four spaces and no two spaces indentations are left.

arnavs:
Yeah, that's what I meant by an aesthetic change. For sure we should do this, but I think it's not worth doing now since it doesn't break the build (?)

Nosferican:
As long as the indentation at least is correct we can push it for later. If the indentation is wrong, we should fix that now.

arnavs:
I think stuff that's supposed to be indented is indented

Nosferican:
I mean we don't have any cells with
```
  a = 1:3
  b = -a
```
instead of 
```
a = 1:3
b = -a
```

PR #278, opened 2018-11-09
Remove Horizontal Scrolling

XiaojunGuan:

Issue #228, opened 2018-10-30
Redo Shortest Path with LightGraphs


jlperla:
- Can add it to the more rich set of packages, should have easy ways to traverse, etc.
- Has `gplot` 
- Could load the graph from a file, or just display a simple one inline rather than the fancy one.
- Can use the fixedpoint

jlperla:
Should then point people to https://juliagraphs.github.io/LightGraphs.jl/latest/pathing.html#General-properties-of-shortest-path-algorithms-1  as an example of "Look for Better algorithms in other packages!"

PR #283, opened 2018-11-11
Remove scrolling

arnavs:
@XiaojunGuan I'm just opening this PR so I can deal with some merge conflicts etc. vs master. Thanks again for all your hard work on this issue #278.

### DIFF
--- a/rst_files/career.rst
+++ b/rst_files/career.rst
@@ -220,7 +220,7 @@ The code for solving the DP problem described above is found below:
       # new life. This is a function of the distribution parameters and is
       # always constant. No need to recompute it in the loop
       v3 = (cp.G_mean + cp.F_mean .+ cp.β .*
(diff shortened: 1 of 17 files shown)


rst_files/short_path.rst
```diff
@@ -158,7 +158,65 @@ Setup
...
+    graph = Dict(zip(0:99, [[(14, 72.21), (8, 11.11), (1, 0.04)],[(13, 64.94), (6, 20.59),
```
jlperla: I think this is the one exception for the scrolling.  Can we revert?  Ultimately in #228 we will want to put this in an external file on the internet to load, but right now I think it looks better if it doesn't take up pages of text.

XiaojunGuan:
Reverted with eb674d7

arnavs:
Thanks @XiaojunGuan, looks good to me and no additional changes (except feedback) since Jesse reviewed so merging now.

PR #291, opened 2018-11-12
edits and typos in getting_started.rst

natashawatkins:
@jlperla I personally find this lecture too technical for `getting_started`, and I'm not really sure we should be advocating that people use cloud-based solutions and docker images when when just beginning to learn

Feel free to disagree with me though, but I think this lecture may be a bit offputting to learners who just want to casually pick up some Julia

Also I think maybe we should install `IJulia` separately, because it will demonstrate the package manager and also make people aware they need to install IJulia to get the kernel... I find I need to reinstall this package often for some reason...

### DIFF
--- a/rst_files/getting_started.rst
+++ b/rst_files/getting_started.rst
@@ -52,7 +52,7 @@ In order of priority, for those just getting started, we suggest
 #. Install the pre-built `docker-based Jupyter <jl_jupyterdocker>`_ from QuantEcon
 #. Install `Anaconda and Julia locally <jl_jupyterlocal>`_ otherwise and manually install QuantEcon based packages
 
-Eventually, you will both want to do a `local installation <jl_jupyterlocal>`_ and move from just using jupyter to using other `tools and editors <tools_editors>`_ such as `Atom/Juno <http://junolab.org/>`_, but don't let the environment get in the way of learning the language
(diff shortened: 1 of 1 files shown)


jlperla:
Is it too technical because it is tough to follow, or because you don't think people should use those technologies?

The docker and web based solutions are, by far, the easiest ones to get started with since the setup process is easier.  Getting Julia installed for non technical users can be very error prone and confusing because there are places to jump between 

Not to mention that a docker or cloud based solution has everything preinstalled and precompiled, so you can get started immediately without 30 minutes of installing and precompiling everything. 

But if you think the docker installation is too tough to follow, we could move it to the local installation section.

natashawatkins:
I think the part that seems too technical is the Docker part... there's a lot of details and a lot of jargon that could be intimidating to people just starting out

jlperla:
Ok, if you think it is too scary we can move it to the local installation section.

jlperla:
See #292

natashawatkins:
What are your thoughts @jstac?

jlperla:
We we move it. The problem with docker is that it is extremely easy to get the whole quantecon setup up and running, but on Windows the docker installation process can sometimes mess up. The whole of these things is to make installing a reproducible setup idiot proof, so if docker doesn't install well it is a problem. Also, hopefully @arnavs can remove further jargon from the installation when it is moved over. I thought I stripped out most of the unnecessary background information, but maybe I didn't go far enough

jstac:
Teaching is hard --- you need to provide enough information to get to the next step but no more, lest you overwhelm the student.  I'll review at some stage but let's all work to minimize jargon and inform on a need-to-know basis.  @natashawatkins has good instincts about this stuff so perhaps we should iterate until she gives thumbs up.

Issue #270, opened 2018-11-08
move deps up to dependency cell 


jlperla:
Right now the `deps.jl` looks something like
```julia
using InstantiateFromURL
activate_github("QuantEcon/QuantEconLecturePackages", tag = "v0.3.1") # activate the QuantEcon environment

using LinearAlgebra, Statistics, Compat # load common packages
```
But for most lectures we want to (`) be able to run the `using` without the instantiation, and (2) add things other than just those common packages on a lecture-by-lecture basis.

So we should
- [x] change the `deps.jl` and `alldeps.jl` to things like
```julia
using InstantiateFromURL
activate_github("QuantEcon/QuantEconLecturePackages", tag = "v0.3.1") ;
```
(note the use of the `;` at the end to prevent the path output from the activate)
Then for every lecture we should:
- [x] Manually add in a code cell underneat the `deps.jl` include that is (on its own with no other code)
```julia
using LinearAlgebra, Statistics, Compat
```
- [x] Move any other `using` up to this cell so that they are all executed at once in the notebook.  The one exception to this is the two introductory sections (i.e. all linked in `index_getting_started_julia` and `index_more_julia` and , where we want to be very careful to help people understand which functions come from which packages)

Also, beyond those three, we should make sure to not have any extra `using` that are not required (e.g. don't go `using QuantEcon` unless we really need a function in it).

jlperla:
Moving this to be considered (partially) for v1.0 as a solution to the precompile output issue.

jlperla:
- [ ] Create `deps_no_using.jl` and `alldepts_no_using.jl`
- [ ] Get rid of the comment text on the using due to scrolling on all of those.
- [ ] Create cell underneath the instantiate with all of the usings in the package, don't need to remove them necessarily (i.e. in the introductory lectures where we say a function is in a particular package).

arnavs:
The commits above were all the ones with compilation issues, so moving this to the `v1.1` milestone.

arnavs:
OK, eliminated all the references to `deps.jl` and `alldeps.jl` (the ones with the included `using LinearAlgebra, Statistics, Compat`), created the standalone code block for that in the lectures, and deleted the old files.

jlperla:
Should this be closed?  If so, lets delete all of the `.jl` files we are not using

arnavs:
Yeah, we can close this. And the `deps.jl` and `alldeps.jl` have bene removed.

PR #294, opened 2018-11-13
remove spaces around tag in deps.jl

natashawatkins:
Is this standard? I'm used to seeing no spaces around keyword arguments

### DIFF
--- a/rst_files/_static/includes/deps.jl
+++ b/rst_files/_static/includes/deps.jl
@@ -1,6 +1,6 @@
 using InstantiateFromURL
 
 # activate the QuantEcon environment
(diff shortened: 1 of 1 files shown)


jlperla:
See #296  On slack there seems to be no consensus, but leaning towards spacing from what I can tell.  I am going to push off the PR until v1.1. discussions.

jlperla:
@arnavs  I suspect that this PR is out of date since we changed over everything in #270  .  Best to choose to either merge or close.

natashawatkins:
I think I'd vote against this now from what I've seen...

arnavs:
Great let’s do it

PR #295, opened 2018-11-13
edits and typos to julia_by_example.rst

natashawatkins:
* I couldn't really see how a setting was changed to use 'Anderson acceleration' - also what is that? An algorithm?
* Might be useful to define machine epsilon

### DIFF
--- a/rst_files/julia_by_example.rst
+++ b/rst_files/julia_by_example.rst
@@ -52,7 +52,8 @@ Set Up
 
 We assume that you've worked your way through :doc:`our getting started lecture <getting_started>` already
 
-In particular, the easiest way to install and precompile all of the Julia packages used in the QuantEcon notes is to go ``] add InstantiateFromURL`` and then work in a Jupyter notebook, as described :ref:`here <jl_jupyter>`
+In particular, the easiest way to install and precompile all of the Julia packages used in the QuantEcon 
+notes is to type ``] add InstantiateFromURL`` and then work in a Jupyter notebook, as described :ref:`here <jl_jupyter>`
 
 
 Other References
@@ -75,8 +76,6 @@ process :math:`\epsilon_0, \epsilon_1, \ldots, \epsilon_T`, where each draw :mat
 .. In other words, we want to generate figures that look something like this:
 .. .. figure:: /_static/figures/test_program_1.png
 ..    :scale: 100%
-
-Fire up a :ref:`Jupyter notebook <jl_jupyter>`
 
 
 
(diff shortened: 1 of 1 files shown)


rst_files/julia_by_example.rst
```diff
@@ -743,7 +737,7 @@ Composing Packages
 
 A key benefit of using Julia is that you can compose various packages, types, and techniques, without making changes to your underlying source
 
-As an example, consider if we want to solve the model with a higher-precision, as floating points cannot be distinguished beyond the machine epsilon for that type,
+As an example, consider if we want to solve the model with a higher-precision, as floating points cannot be distinguished beyond the machine epsilon for that type
```
arnavs:
```suggestion
As an example, consider if we want to solve the model with a higher-precision, as floating points cannot be distinguished beyond the machine epsilon for that type (recall that computers approximate real numbers to the nearest binary  of a given precision; the *machine epsilon* is the smallest nonzero magnitude)

File: rst_files/julia_by_example.rst [hunk 2]
```diff
@@ -704,17 +700,16 @@ But best of all is to avoid writing code altogether
...
 
 The ``fixedpoint`` function from the ``NLsolve.jl`` library implements the simple fixed-point iteration scheme above
 
-Since the ``NLsolve`` library only accepts vector based inputs, we needed to make the ``f(v)`` function broadcast on the ``+`` sign, and pass in the initial condition as a vector of length 1 with ``[0.8]]``
+Since the ``NLsolve`` library only accepts vector based inputs, we needed to make the ``f(v)`` function broadcast on the ``+`` sign, and pass in the initial condition as a vector of length 1 with ``[0.8]``
 
 While a key benefit of using a package is that the code is clearer, and the implementation is tested, by using an orthogonal library we also enable performance improvements
 
-In particular, we can use the ``Anderson acceleration`` with a memory of 5 iterations, by changing a setting
+In particular, we can use the **Anderson acceleration** with a memory of 5 iterations, by changing a setting
```
arnavs: Good point. I think we can just eliminate this, since we updated `NLsolve` to make this the default 

```suggestion
```

natashawatkins:
Ok - could you maybe indent the string on the following line so that it aligns with the start of the string?

rst_files/julia_by_example.rst
```diff
@@ -659,34 +657,32 @@ To enable this, Julia has two features:  named function parameters, and named tu
...
     # define a map and parameters
     p = 1.0
     β = 0.9
     f(v) = p + β * v # note that p and β are used in the function!
 
-    sol = fixedpointmap(f, iv = 0.8, tolerance = 1.0E-8) # don't need to pass 
-    println("Fixed point = $(sol.value), and |f(x) - x| = $(sol.normdiff) in $(sol.iter)"*
-    "iterations")
+    sol = fixedpointmap(f, iv=0.8, tolerance=1.0E-8) # don't need to pass 
+    println("Fixed point=$(sol.value), and |f(x) - x| = $(sol.normdiff) in $(sol.iter) iterations")
```
arnavs:
```suggestion
    println("Fixed point = $(sol.value), and |f(x) - x| = $(sol.normdiff) in $(sol.iter)"*
               " iterations")
```

arnavs:
OK, made these changes, merging now. Thanks @natashawatkins

PR #297, opened 2018-11-13
edits and typos in julia_essentials.rst

natashawatkins:
* I think generally for readability we should try to align code comments and add more line breaks between functions etc.
* I added in `\approx` demo because I thought that was cool in the example lecture
* I think there was one point where an error was meant to be displayed? I think we are using `skip-test` for this, not `no-execute`
* I wasn't sure what one sentence meant

### DIFF
--- a/rst_files/julia_essentials.rst
+++ b/rst_files/julia_essentials.rst
@@ -558,12 +558,15 @@ For "not equal" use ``!=`` or ``≠`` (``\ne<TAB>``)
 
     x != 3
     
-Julia can also test approximately equal with ``≈`` (``\approx<TAB>``)
+Julia can also test approximate equality with ``≈`` (``\approx<TAB>``)
 
 
 .. code-block:: julia
 
     1 + 1E-8 ≈ 1
+
+Be careful when using this, however, as there are subtleties involving the scales of the quantities compared
+
 
 .. I think this is a a bad practive
 .. We can chain inequalities:
@@ -750,9 +753,9 @@ For example, in the call
 
 .. code-block:: julia
 
-    f(x; a = 1) = exp(cos(a * x))  # Note the ; in the definition
+    f(x; a = 1) = exp(cos(a * x))  # note the ; in the definition
 
-    f(pi, a = 2)                   # Calling with ; is usually optional and generally discouraged
+    f(pi, a = 2) # calling with ; is usually optional and generally discouraged
 
 
 Broadcasting
@@ -800,7 +803,7 @@ In doing this we'll exploit the fact that, if we take ``k`` independent standard
     function chisq(k)
         @assert k > 0
         z = randn(k)
-        return sum(z -> z^2, z)  # Same as `sum(x^2 for x in z)`
+        return sum(z -> z^2, z)  # same as `sum(x^2 for x in z)`
     end
 
 The macro ``@assert`` will check that the next expression evaluates to ``true``, and will stop and display an error otherwise
@@ -832,7 +835,7 @@ The broadcasting notation is not simply vectorization, as it is able to "fuse" m
     x = 1.0:1.0:5.0
     y = [2.0, 4.0, 5.0, 6.0, 8.0]
     z = similar(y)
-    z .= x .+ y .- sin.(x)  # Generates efficient code instead of many temporaries
+    z .= x .+ y .- sin.(x) # generates efficient code instead of many temporaries
 
 A convenience macro for adding broadcasting on every function call is `@.`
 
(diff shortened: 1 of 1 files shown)


jlperla:
@natashawatkins all of your comments are great, but...

> I think generally for readability we should try to align code comments and add more line breaks between functions etc.

That is outside of the style guide https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md (and generally inconsistent with julia comment styling).

If you disagree, this can be discussed but I would prefer not to make a big change in the comments when we have already started to move things to be more consistent.

Is there any way you could revert those changes in your PRs to make it easier for @arnavs  to merge?

>  I think we are using skip-test for this, not no-execute

How does the `skip-test` work?  Is this just for the jupinx testing?  If so, then I think there are a whole bunch of places we could consider this.

natashawatkins:
`skip-test` is meant to produce the error and display it in the lecture, but not count it towards the lecture passing status
`no-execute` is for code that should be skipped, like pseudo-code or a template of code etc.

rst_files/julia_essentials.rst
```diff
@@ -851,38 +861,43 @@ Another place that you may use a ``Ref`` is to fix a function parameter you do n
...
+    f.(Ref([3, 4, 5]), [2, 3])    # Broadcasting over 2nd parameter, fixing first
 
 
 Scoping and Closures
 =====================
 
 Since global variables are usually a bad idea, we will concentrate on understanding the role of good local scoping practice
 
-That said, many of the variables in a Jupyter notebook are global, even the whole text could be copied into a function
+That said, global variables are often convenient and easy to use in Jupyter notebooks
```
jlperla: This is missing the spirit of the comment (which is: you can copy the whole thing inside of a function and they cease to be global).  But the comment is very unclear.  How about something like this instead (which you could edit and merge)
```
That said, while many of the variables in these Jupyter notebook are global, we have been careful to write the code so that the entire code could be copied inside of a function

When copied inside a function, variables become local and functions become closures
```

natashawatkins:
Are you sure that's not Julia style? I see an instance of aligning comments here: https://docs.julialang.org/en/v1/manual/style-guide/index.html#Avoid-writing-overly-specific-types-1

I personally found the comments difficult to follow as they were

There was also some inconsistency with capitlisation and putting a space following the hashtag

jlperla:
> Are you sure that's not Julia style? I see an instance of aligning comments here: https://docs.julialang.org/en/v1/manual/style-guide/index.html#Avoid-writing-overly-specific-types-1

I don't think there is any standardization of any of this, but more often than not they aren't aligned.

The bigger question is: if we start aligning all comments will it look odd if we align none of them?  If this isn't something you are worried about, then I don't especially care.  If you want to align comments in some of these, go for it. 

> There was also some inconsistency with capitlisation and putting a space following the hashtag

Yes, older code has only been partially converted over. This one would take time and was lower priority.

jlperla:
I made an issue for the `no-test` thing: #300  

Feel free to make that change if you come across any where it makes sense (as long as you check that the jupyter notebook looks good after the change).  There may be a few places where we used a no-execute because the alternatives were worse.  One of them is the `?` docstring lookup, which looks terrible if we show the raw output.

rst_files/julia_essentials.rst
```diff
@@ -974,29 +990,30 @@ While the above was convenient, there are other times when you want to simply fi
...
+        f(x) = a * x^2  # Refers to the `a` passed in the function
+        f(1)            # Univariate function
     end
     g(0.2)
 
 Comparing the two:  the key here is not that ``a`` is a global variable, but rather that the ``f`` function is defined to capture a variable from an outer scope
 
 This is called a **closure**, and are used throughout the lectures
 
+**what does this mean**
```
arnavs: Changed this to `captured variable`, which is what I think was meant. Also, important to make sure comments are left as genuine rST comments, since output like this can sneak through.

arnavs: (That is, if comments are left at all, per the other PR).

jlperla: Wait, you got rid of **closure**?  That is an important word to teach people.  A closure also "captures a variable", it isn't a captured variable?

arnavs: No, didn't get rid of that section. Just changed "closure variable" to "captured variable" in: 

```
This is called a **closure**, and are used throughout the lectures

It is generally bad practice to modify the closure variable in the function, but otherwise the code becomes very clear

One place where this can be helpful is in a string of dependent calculations
```

jlperla: So you changed it from "modify the closure variable in the function" to "modify the captured variable in the function"?  Sounds good.

arnavs: Exactly.

PR #299, opened 2018-11-13
edits and typos in introduction_to_types.rst

natashawatkins:


rst_files/introduction_to_types.rst
```diff
@@ -171,12 +175,14 @@ While you will never have a ``Real`` number directly in memory, the abstract typ
 .. But we've also seen references to types such as ``Real`` and ``AbstractFloat``
 
 
+**can we think of a different subheading here, it's the same as above**
```
jlperla: I think the subheading here could just be removed?

jlperla:
@natashawatkins As with the other one, I don't want to change the capitalization of comments if we can avoid it.  If you feel very strongly about the alignment of the comments, I can live with it.

jlperla:
I pinged you on slack to discuss, but I think we we need to be careful putting comments in the `rst` source for discussion.  If we do, then some will inevitably sneak through and get into the output.

Instead, maybe you make what you think is a reasonable, but minimal change that would look good in the output, and then put in a comment in the PR to discuss?

@arnavs  Make sure you scan for those comments before merging any PRs.

rst_files/introduction_to_types.rst
jlperla: Sorry, I looked at the source again for this and take back my comment.

@arnavs @natashawatkins  Thoughts on calling this subsection `Exploring the Type Hierarchy`?

rst_files/introduction_to_types.rst
```diff
@@ -665,14 +706,22 @@ The best resource is to carefully read other peoples code, but a few sources to
 
 Now why would we emphasize naming as style as a crucial part of the lectures?
 
-Because it is an essential tool for creating research that is **reproducible** and `**correct** <https://en.wikipedia.org/wiki/Correctness_(computer_science)>`_
+Because it is an essential tool for creating research that is 
+**reproducible** and `**correct** <https://en.wikipedia.org/wiki/Correctness_(computer_science\)>`_
```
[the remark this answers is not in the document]
jlperla: Ah Sorry I see that it is being escaped now.  It looks to me like the first `(` is not escaped.  i.e. should be `https://en.wikipedia.org/wiki/Correctness_\(computer_science\)`

rst_files/introduction_to_types.rst
arnavs: Since the whole section is about types like `Number` and `Int64`, how about something like `Example: Numerical Types`?

rst_files/introduction_to_types.rst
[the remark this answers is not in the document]
arnavs: Done

rst_files/introduction_to_types.rst
```diff
@@ -171,12 +175,14 @@ While you will never have a ``Real`` number directly in memory, the abstract typ
 .. But we've also seen references to types such as ``Real`` and ``AbstractFloat``
 
 
+**can we think of a different subheading here, it's the same as above**
+
 The Type Hierarchy
```
arnavs:
```suggestion
Subtypes and Supertypes
```

rst_files/introduction_to_types.rst
```diff
@@ -364,17 +376,17 @@ Syntax for Creating Concrete Types
...
 
 
 Let's start with a trivial example where the ``struct`` we build has fields named ``a, b, c``, are not typed
 
 .. code-block:: julia
 
-    struct FooNotTyped # immutable by default, use `mutable struct` otherwise 
-        a # BAD! Not typed
+    struct FooNotTyped  # Immutable by default, use `mutable struct` otherwise 
+        a               # BAD! Not typed
```
arnavs: @jlperla Just wanted to check about this. The convention we've been using in the rest of the lectures is to avoid explicit type constraints on the fields of an immutable struct (to make them closer to `NamedTuples`s, and to reduce the quantity of type annotations in general).

jlperla: We should *never* have structs without type declartions.

jlperla: It is a performance disaster.  This is part of the reason that we should basically never have structs.

Nosferican: Can do
```
struct A{T1, T2, T3}
    x::T1
    y::T2
    z::T3
end
```
might be good to have as the fix.

jlperla: @Nosferican  Do you mean in the current lectures that we haven't been able to rip the structs out of?  Yes, that is the only solution for now.

As for the notes themselves, if you look at http://compare-lectures.quantecon.org/jl/introduction_to_types.html# you will see that is exactly what we show afterwards.

arnavs: Makes sense --- for some reason I figured the compiler would silently parametrize with the field parameters like it does for the `NamedTuple`. 

While we're waiting on build stuff, we can pay extra attention to the `struct` issue.

arnavs:
Merging after incorporating feedback from Jesse above.

PR #330, opened 2018-11-17
further edits to getting_started.rst

natashawatkins:
This is reading a lot better in my opinion

@jstac I was wondering whether you thought it would be a good idea to mention QuantEcon Notes in the "sharing Jupyter notebooks" section?

I fixed some typos etc. and added in links to things not previously mentioned ie. Anaconda and IJulia


natashawatkins:
Let's add in a mention of QuantEcon notes in the sharing section

PR #338, opened 2018-11-17
Update optimization_solver_packages.rst

gava31:
Change in line 26: there are four methods. Other harmless changes.


arnavs:
Sure, looks good to me.

PR #343, opened 2018-11-18
Update testing.rst

pedropessoa:
Typo: "one" in line 753 misplaced?


arnavs:
Yeah, the “one” is more informal here than we should probably be

PR #353, opened 2018-11-19
Unshown equations

Qirui-UBC:
In line 132-135 and line 139-142, the equations are not shown on the reading mode. I tried to remove the label and consequently rephrase the sentence in line 170.


arnavs:
Hey @Qirui-UBC, this would ordinarily be a great edit, but this is something that should work in principle and we're ironing it out on the build side. So going to close, but this is a solid PR with a parseable, minimal fix.

PR #356, opened 2018-11-19
Just for assignment 7

AAdibnia:
In one of the lines you need to use ';'.


jlperla:
Thanks @AAdibnia   I will close because I think you had done a fork on an earlier version, and some crazy stuff happened!

AAdibnia:
Hi Jessi,

Sorry about that, I am wondering is it enough for part 2a and 2b, or should
I do more?

Thanks
Ali

jlperla:
That is fine.  But if you ever see hundreds of files modified in a PR, then you have done somehting wrong (e.g. chosen to make changes to a particular branch, etc.)

AAdibnia:
Oh, I see thanks for this information.

PR #357, opened 2018-11-19
Missing comma

AliKarimirad:
There is a missing comma. I edit it
Exercise 2.b by using fork


arnavs:
Hey @AliKarimirad, I think this edit makes sense but the style of these lectures (e.g. the two sentences on either side) is to keep punctuation to a minimum and use direct sentences. So unless we need to set apart two distinct ideas, probably should not use the comma.

PR #371, opened 2018-11-19
Rewrite Lake Model

arnavs:
Still need to iron out the connection with the McCall.


rst_files/lake_model.rst
```diff
@@ -220,58 +220,26 @@ Here's the code:
...
-        x = 0.5 * ones(2)
-        error = tol + 1
-        while (error > tol)
-            new_x = lm.A_hat * x
-            error = maximum(abs, new_x - x)
-            x = new_x
-        end
-        return x
-    end
+    rate_steady_state(lm) = fixedpoint(x -> lm.A_hat * x, [0.5, 0.5], inplace = false).zero
```
Nosferican: Should this have a convergence check?

rst_files/lake_model.rst
```diff
@@ -220,58 +220,26 @@ Here's the code:
...
 
     function simulate_stock_path(lm, X0, T)
-        X_path = zeros(eltype(X0), 2, T)
-        X = copy(X0)
-        for t in 1:T
-            X_path[:, t] = X
-            X = lm.A * X
-        end
-        return X_path
+        raw = [lm.A^t * X0 for t in 0:T-1]
```
Nosferican: Why not broadcast this line?

rst_files/lake_model.rst
```diff
@@ -220,58 +220,26 @@ Here's the code:
...
+        reshaped = hcat(raw...)' # each row is one time period
```
Nosferican: I believe `reduce(hcat, raw)'` avoids the splatting.

rst_files/lake_model.rst
```diff
@@ -514,12 +485,13 @@ Let's plot the path of the sample averages over 5,000 periods
     using QuantEcon, Roots, Random
 
     Random.seed!(42)
```
Nosferican: Moved to the test block as well as the `using Test, Random`

rst_files/lake_model.rst
jlperla: We should avoid `reduce(hcat` since it is very unclear.  Why not just make a small funciton and broadcast for something so simple?

rst_files/lake_model.rst
```diff
@@ -541,7 +514,7 @@ Let's plot the path of the sample averages over 5,000 periods
   :class: test
 
   @testset begin
-      @test xbar[1] ≈ 0.043921027960428106
+      @test round(xbar[1], digits = 5) == round(0.043921027960428106, digits = 5)
```
Nosferican: I think here is best to use the `atol = 1e-5`

rst_files/lake_model.rst
jlperla: Also we can remove the inplace, right?

rst_files/lake_model.rst
```diff
@@ -669,84 +641,114 @@ Following :cite:`davis2006flow`, we set :math:`\alpha`, the hazard rate of leavi
...
+
+    using Distributions, LinearAlgebra, Compat, Expectations, Parameters, NLsolve, Plots
+
+    function solve_mccall_model(mcm; U_iv = 1.0, V_iv = ones(length(mcm.w)), tol = 1e-5, 
+                                iter = 2_000)
+        # α, β, σ, c, γ, w = mcm.α, mcm.β, mcm.σ, mcm.c, mcm.γ, mcm.w
+        @unpack α, β, σ, c, γ, w, dist, u = mcm 
+        
+        # parameter validation
+        @assert c > 0.0
```
Nosferican: Better to use `throw(ArgumentError(msg))`, but maybe best would be to use ArgCheck.jl

rst_files/lake_model.rst
```diff
@@ -669,84 +641,114 @@ Following :cite:`davis2006flow`, we set :math:`\alpha`, the hazard rate of leavi
...
+        # Bellman operator T. Fixed point is x* s.t. T(x*) = x*
+        function T(x)
+            V = x[1:end-1]
+            U = x[end]
+            [u_w + β * ((1 - α) * V .+ α * U); u_c + β * (1 - γ) * U + β * γ * E * max.(U, V)]
+        end 
+
+        # value function iteration  
+        x_iv = [V_iv; U_iv] # initial x val
+        xstar = fixedpoint(T, x_iv, iterations = iter, xtol = tol).zero
```
Nosferican: Same on the convergence check.

rst_files/lake_model.rst
jlperla: More importantly, why do we need to round at all?  This should be deterministic so that we can see if anything changed.  And if there is something stochastic in the test, then we should have it seeded to be deterministic anyways.

rst_files/lake_model.rst
jlperla: No, the style guide says use `@assert` unless you are actually doing a proper error handling.  This is a test of an invariant.

rst_files/lake_model.rst
```diff
@@ -669,84 +641,114 @@ Following :cite:`davis2006flow`, we set :math:`\alpha`, the hazard rate of leavi
...
+
+        # value function iteration  
+        x_iv = [V_iv; U_iv] # initial x val
+        xstar = fixedpoint(T, x_iv, iterations = iter, xtol = tol).zero 
+        V = xstar[1:end-1]
+        U = xstar[end]
+        
+        # compute the reservation wage 
+        wbarindex = searchsortedfirst(V .- U, 0.0)
+        if wbarindex >= length(w) # if this is true, you never want to accept
```
Nosferican: Could be an `ifelse`

rst_files/lake_model.rst
```diff
@@ -669,84 +641,114 @@ Following :cite:`davis2006flow`, we set :math:`\alpha`, the hazard rate of leavi
...
+
+.. code-block:: julia
 
-The second piece of code repeated from :doc:`the McCall model lecture <mccall_model>` is used to complete the reservation wage
+    # a default utility function
+    u(c, σ) = (c^(1 - σ) - 1) / (1 - σ) 
 
+    # model constructor
+    McCallModel = @with_kw (α = 0.2, 
+        β = 0.98, # discount rate
```
Nosferican: Align these.

rst_files/lake_model.rst
jlperla: Is there a benefit in readability or consistency with using `ifelse` there?

Nosferican: For a simple `ifelse` is easier to read
```
x = ifelse(cond, val_true, val_false)
```
vs
```
if cond
    x = val_true
else
    x = val_false
end
```

rst_files/lake_model.rst
```diff
@@ -220,58 +220,26 @@ Here's the code:
...
+        A = [(1-λ) * (1-d) + b  (1-d) * α + b; # carry out intermediate calculations
             (1-d) * λ          (1-d) * (1-α)]
         A_hat = A ./ (1 + g)
-
-        return LakeModel(λ, α, b, d, g, A, A_hat)
+        return (λ = λ, α = α, b = b, d = d, g = g, A = A, A_hat = A_hat) # return a NamedTuple
     end
 
-    function rate_steady_state(lm, tol = 1e-6)
-        x = 0.5 * ones(2)
```
Nosferican:
```
x = fill(0.5, 2)
```

rst_files/lake_model.rst
Nosferican: Goes against the principle of using standard functions rather than defining your own, but how about
```
reduce((x, y) -> [x y], obj)
```

rst_files/lake_model.rst
jlperla: So in this particular example
```julia
    if wbarindex >= length(w)
        w_bar = Inf
    else
        w_bar = w[wbarindex] # otherwise, return the number
    end
```
becomes

```
 # if index doesn't exist, then w_bar = Inf since they never accept
w_bar = ifelse(wbarindex >= length(w), Inf, w[barindex])
```
I am not sure what is more readable for intro users.  Close call.

rst_files/lake_model.rst
jlperla: That is impenetrable for someone just starting to program.

`reduce` is effectively banned in https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md outside of parallel programming/etc.

It is too difficult for intro users to understand.  Look at the example there.  using `loops` with a little more code is easier for them to understand.

rst_files/lake_model.rst
Nosferican: Personally, I would use
```
w_bar = wbarindex ≥ length(w) ? Inf : w[wbarindex]
```
since we lifted the ban on the ternary operator and seems pretty easy to follow.

rst_files/lake_model.rst
Nosferican: Then it might might need a re-write of the structured used in the lecture.

rst_files/lake_model.rst
jlperla: When did we lift the ternary ban?  The current guideline is identical to before
> Minimize use of the ternary operator. It is confusing for new users, so use it judiciously, and never purely to make code more terse.

This is a case where the ternary operator may be relatively readable since it follows a simple pattern... but we should always be suspicious.  It is worst when used nested in other functions.  But if we move to this, we need a comment explaining the ternary operator (each time we use it, since people will read these thigns out of orderr).

All that said, the `if` approach is so clear that I don't see the benefit of changing things like this.  Our time is spent worrying about first-order readability changes instead of wondering if we can make it a little more concise at the cost of using an operator we need to explain each time we use it.

rst_files/lake_model.rst
jlperla: Sure.  Some simple for loops putting things in a matrix might make this very clear.  As long as we move away from a functional programming style for it, I am happy.

rst_files/lake_model.rst
Nosferican: I thought we had allowed it only for simple cases. Guess I was wrong. It is pretty common for languages economists might use. MATLAB, R, Python, Stata, and SAS have it. Nvm.

rst_files/lake_model.rst
arnavs: Both are true.

rst_files/lake_model.rst
arnavs: Just to be clear, there wasn't an issue with _ever_ seeding the RNG. The issue is just that we don't want that thing to pop up before every block. So I think we can leave this as is.

Nosferican: If used in the lectures then it's fine. Just split the `using` statement and the seeding can go right before the execution cell to ensure it gets seeded when it should.

arnavs: Roger that.

rst_files/lake_model.rst
arnavs: Done.

rst_files/lake_model.rst
arnavs: Thanks, done.

rst_files/lake_model.rst
arnavs: I think this is fine as-is.

rst_files/lake_model.rst
arnavs: OK, done. Will need to check that it doesn't require scrolling.

rst_files/lake_model.rst
arnavs: This comment applies to the old code, not the code in the PR

rst_files/lake_model.rst
arnavs: We can definitely move away from the splatting, but I don't think that's a pressing concern. It seemed that @jlperla agreed, but let me know if I'm incorrect.

jlperla: No I hate the splatting.  Pretty much anytime the word `hcat` or `reduce` or `...` come up I am bound to hate it.

What is wrong with a for loop here?  Everybody can understand the loop equivalent to replace
```
        raw = [lm.A^t * X0 for t in 0:T-1]
        reshaped = hcat(raw...)' # each row is one time period
```
In fact, you can probably do it faster as well by going backwards in the iteration.  Something like
```
vals = # PREALLOCATE
vals[1,:] = X0
for t in 2:T
   vals[t, :] = lm.A * X
```
or something like that

arnavs: That works too, will do.

arnavs:
I've been wrangling with this for a while but it keeps failing for one reason or another. Assigning over to @Nosferican who will have time tomorrow.

Highlights for Jose:

* For now, don't worry about any unit tests apart from the McCall integration. 

* I adapted the McCall struct to take a tax level, which we use to shift the utility function args (i.e., `u_c = u(c - t)`. This way, we don't run afoul of the expectations operator by sometimes overstepping distribution bounds. 

* I would see some weird grid effects, where if I changed the step size, things which weren't able to converge magically converged. Note that we moved away from the `QuantEcon` brent to `find_zero` in `Roots.jl`

arnavs:
Closing this per Slack @Nosferican

PR #373, opened 2018-11-19
Qirui's change

Qirui-UBC:


arnavs:
Hi @Qirui-UBC, thanks for the change. Some points: 

* This PR still has the commit which removes the equation labels we want to hold on to. It's important to remember that PRs are time-dependent --- they'll give the upstream the whole history, and any new commits you push in the future. 

* Try to name your branches with alphanumerics, dashes, and underscores only. Characters like `'` sometimes need to be escaped and are hard to parse.

PR #379, opened 2018-11-20
Pooya

PooyaFa:


rst_files/data_statistical_packages.rst
```diff
@@ -93,7 +93,7 @@ Named tuples can also be used to construct a DataFrame, and have it properly ded
 Working with Missing
 -----------------------
 
-As we discussed in `fundamental types <missing>`_, the semantics of ``missing`` are that mathematical operations will not silently ignore it
+As we discussed in `fundamental types <fundamental_types>`_, the semantics of ``missing`` are that mathematical operations will not silently ignore it
```
arnavs:
```suggestion
As we discussed in `fundamental types <missing>`_, the semantics of ``missing`` are that mathematical operations will not silently ignore it
```

I can see why this is confusing, but the relevant section is actually called `missing` (i.e., "the semantics of `missing`...")

PR #383, opened 2018-11-20
Update general_packages.rst

yywingliang:
Proofs of concept, no dash in between for plural form.
I guess "it is finding ... using FastGaussQuadrature"; there should be no "and".


arnavs:
I think the dash is technically correct here even for the plural (e.g., `secretaries-general`), but it's probably easier to read without it. Thanks for the PR

PR #384, opened 2018-11-20
edits to generic_programming.rst

natashawatkins:
@jlperla Personally I feel like this lecture could be cut down and more succinct - there's a lot of detail, possibly too much detail

There seems to be an issue with links that contain `(`.. I haven't fixed this

Why are mathematical terms like Field or Ring being capitalised?

### DIFF
--- a/rst_files/generic_programming.rst
+++ b/rst_files/generic_programming.rst
@@ -575,7 +575,7 @@ Reals and Algebraic Structures
 Thinking back to the mathematical motivation, a `Field <https://en.wikipedia.org/wiki/Field_\(mathematics\)>`_ is an `Ring` with a few additional properties, among them
 
     * a multiplicative inverse: :math:`a^{-1}`
-    * an inverse operation for multiplication: :math:`\frac{a}{b} = a \cdot b^{-1}`
+    * an inverse operation for multiplication: :math:`a / b = a \cdot b^{-1}`
 
 Furthermore, we will make it a `Total Ordered <https://en.wikipedia.org/wiki/Total_order#Strict_total_order>`_ Field with
 
--- a/rst_files/generic_programming.rst
+++ b/rst_files/generic_programming.rst
@@ -572,7 +572,7 @@ Hopefully this showcases the power of generic programming:  with a well-designed
 Reals and Algebraic Structures
 =======================================
  
-Thinking back to the mathematical motivation, a `field <https://en.wikipedia.org/wiki/field_\(mathematics\)>`_ is an ``ring`` with a few additional properties, among them
+Thinking back to the mathematical motivation, a `field <https://en.wikipedia.org/wiki/field_\(mathematics\)>`_ is a ``ring`` with a few additional properties, among them
 
     * a multiplicative inverse: :math:`a^{-1}`
     * an inverse operation for multiplication: :math:`a / b = a \cdot b^{-1}`
@@ -1003,7 +1003,8 @@ On the other hand, if we use change the function to return ``0`` if `x <= 0`, it
     @code_warntype f(1.0)
 
 The solution is to use the ``zero(x)`` function which returns the additive identity element of type ``x``
-n the other hand, if we change the function to return ``0`` if ``x <= 0``, it is type-unstable with  floating point
+
+On the other hand, if we change the function to return ``0`` if ``x <= 0``, it is type-unstable with  floating point
 
 .. code-block:: julia
 
--- a/rst_files/generic_programming.rst
+++ b/rst_files/generic_programming.rst
@@ -27,26 +27,26 @@ In this lecture we delve more deeply into the structure of Julia, and in particu
 
 Understanding them will help you
 
-* Design code that matches the "white-board" mathematics
+* design code that matches the "white-board" mathematics
 
-* Create code that can use (and be used) by a variety of other packages
+* create code that can use (and be used by) a variety of other packages
 
-* Write "well organized" Julia code that's easy to read, modify, maintain and debug
+* write "well organized" Julia code that's easy to read, modify, maintain and debug
 
-* Improve the speed at which your code runs
+* improve the speed at which your code runs
 
 (Special thank you to Jeffrey Sarnoff)
 
 Generic Programming is an Attitude
 -----------------------------------------------
 
-From ``Mathematics to Generic Programming`` :cite:`stepanov_mathematics_2014`
+From *Mathematics to Generic Programming* :cite:`stepanov_mathematics_2014`
 
     Generic programming is an approach to programming that focuses on designing algorithms and data structures so that they work in the most general setting without loss of efficiency... Generic programming is more of an *attitude* toward programming than a particular set of tools.
 
 In that sense, it is important to think of generic programming as an interactive approach to uncover generality without compromising performance -- and not a set of rules to decomposing taxonomies of abstractions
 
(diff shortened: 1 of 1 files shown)


jlperla:
Thanks Natasha. I will explain more later why the content in this section (which you will notice is in the advanced Julia section of the notes) is so essential.

One thing that we could do, though, is move the "understanding the compiler" section to the "need for speed". That was older material from the notes that I thought was useful, but was too advanced for the introduction to Julia section. Is that a good idea?

rst_files/generic_programming.rst
```diff
@@ -564,7 +575,7 @@ Reals and Algebraic Structures
 Thinking back to the mathematical motivation, a `Field <https://en.wikipedia.org/wiki/Field_\(mathematics\)>`_ is an `Ring` with a few additional properties, among them
 
     * a multiplicative inverse: :math:`a^{-1}`
-    * an inverse operation for multiplication: :math:`a / b = a \cdot b^{-1}`
+    * an inverse operation for multiplication: :math:`\frac{a}{b} = a \cdot b^{-1}`
```
jlperla: This is only true for commutative fields, so I think we should change it back.

rst_files/generic_programming.rst
```diff
@@ -713,12 +724,15 @@ You will notice that types in Julia represent a tree with ``Any`` at the root
 
 The tree structure has worked well for the above examples, but it doesn't allow us to associate multiple categorizations of types
 
-For example, a semi-group type would be useful for a writing generic code (e.g. continuous-time solutions for ODEs and matrix-free methods), but cannot be implemented rigorously since the ``Matrix`` type is a semi-group as well as an ``AbstractArray``, but not not all ``semi-groups" are ``AbstractArray``
+For example, a semi-group type would be useful for a writing generic code (e.g. 
+continuous-time solutions for ODEs and matrix-free methods), but cannot be 
+implemented rigorously since the ``Matrix`` type is a semi-group as well 
+as an ``AbstractArray``, but not all ``semi-groups`` are ``AbstractArray``
```
jlperla: The problem is that ``semi-group`` is not a defined structure in the libraries we use, so we should instead change it to either "semi-group" or perhaps just drop the quotes around semi-group

jlperla:
@arnavs  See my few comments, but I think we can basically merge this with the minor change.

A few other thigns you could look for before merging
- [ ] "Why are mathematical terms like Field or Ring being capitalised?"  They shouldn't be given the current style standards.  Arnav, can you strip those capitlizations?
- [ ] I don't know the latest on the links with `(` in them.  Arnav, is this currently broken in the HTML or notebook output?  If it isn't supposed to be broken, you could see if there are other issues
- [ ] After this is merged, I will move the "understanding the compiler" section to the "need for speed" lectures.

rst_files/generic_programming.rst
arnavs: I thought fields were always commutative... i.e., the elements are an abelian group under addition, and the units/invertibles are an abelian group under multiplication

arnavs: http://mathworld.wolfram.com/FieldAxioms.html

jlperla: Sorry, my brain was mixed up with the semi-group below.

Never theless, the ambigurity of the fraction means that I think `a / b` is preferrable in this instance, even if `b \ a = frac{a}{b}` would be equivalent

arnavs: All good, will revert the change.

rst_files/generic_programming.rst
arnavs: Dropped the backticks.

arnavs:
Looks good to me. Fixed a typo and made sure the links work. (@natashawatkins, can't reproduce what you were seeing- the escaped parentheses seem to work fine). 

Will merge now

arnavs:
The issue with the two links on the HTML is just that they're cut off prematurely; i.e., the final `)` is not there. I thought this was a typo on our part, but maybe it's not. Will file an issue with Matt.

jlperla:
Using the merged version, I just made a few small edits (i.e. I had forgotten to strip a section of links on OO programming out) and moved the compiler stuff to need for speed.

natashawatkins:
Ahh I forgot to click 'submit review' for my last two comments, see above

rst_files/generic_programming.rst
```diff
@@ -27,26 +27,26 @@ In this lecture we delve more deeply into the structure of Julia, and in particu
...
 
-From ``Mathematics to Generic Programming`` :cite:`stepanov_mathematics_2014`
+From *Mathematics to Generic Programming* :cite:`stepanov_mathematics_2014`
 
     Generic programming is an approach to programming that focuses on designing algorithms and data structures so that they work in the most general setting without loss of efficiency... Generic programming is more of an *attitude* toward programming than a particular set of tools.
 
 In that sense, it is important to think of generic programming as an interactive approach to uncover generality without compromising performance -- and not a set of rules to decomposing taxonomies of abstractions
 
-As we will see, the core approach is to treat data-structures and algorithms as loosely coupled, and is in direct contrast to the "isa" approach of object-oriented programming
+As we will see, the core approach is to treat data structures and algorithms as loosely coupled, and is in direct contrast to the "isa" approach of object-oriented programming
```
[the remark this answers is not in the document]
jlperla: Good point.  maybe change the "is-a" to
```
`is-a <https://en.wikipedia.org/wiki/Is-a>`_ 
```
i.e., hiearchical taxonomies  as opposed to https://en.wikipedia.org/wiki/Has-a which you can also do in OOP, but it is not the prevailing approach.

rst_files/generic_programming.rst
```diff
@@ -27,26 +27,26 @@ In this lecture we delve more deeply into the structure of Julia, and in particu
...
 
 Generic Programming is an Attitude
 -----------------------------------------------
 
-From ``Mathematics to Generic Programming`` :cite:`stepanov_mathematics_2014`
+From *Mathematics to Generic Programming* :cite:`stepanov_mathematics_2014`
 
     Generic programming is an approach to programming that focuses on designing algorithms and data structures so that they work in the most general setting without loss of efficiency... Generic programming is more of an *attitude* toward programming than a particular set of tools.
 
 In that sense, it is important to think of generic programming as an interactive approach to uncover generality without compromising performance -- and not a set of rules to decomposing taxonomies of abstractions
```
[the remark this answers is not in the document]
jlperla: I guess the question is how much we want to target people with previous OO knowledge.  The battle I have had is that people start making Julia code look like Python of non-generic C++, which is a recipe for disaster.  If you never learned OO in sufficient detail, then count yourself lucky  :-)

Let me put in an additional statement there to make it more clear.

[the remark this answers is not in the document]
jlperla: Master now has an updated version of these.

rst_files/generic_programming.rst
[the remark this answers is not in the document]
natashawatkins: I don't actually know what 'taxonomy' means, or 'taxonomies of abstractions' - is there another way to phrase these sentences?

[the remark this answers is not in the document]
jlperla: I removed taxonomies of abstractsion in that location.

rst_files/generic_programming.rst
```diff
@@ -1197,7 +1213,7 @@ n the other hand, if we use change the function to return ``0`` if `x <= 0`, it
 .. _spec_field_types:
 ..
 ..
-.. Specifying Field Types
+.. Specifying field Types
```
arnavs: Probably one victim of a global find/replace

PR #386, opened 2018-11-21
edits to data_statistical_packages.rst

natashawatkins:
* I think the lecture could have a more specific overview or a summary of what's covered in the lecture

### DIFF
--- a/rst_files/data_statistical_packages.rst
+++ b/rst_files/data_statistical_packages.rst
@@ -27,11 +27,12 @@ Setup
 DataFrames
 ========================
 
-A useful package for working with data is `DataFrames <https://github.com/JuliaStats/DataFrames.jl>`_
+A useful package for working with data is `DataFrames.jl <https://github.com/JuliaStats/DataFrames.jl>`_
 
 The most important data type provided is a ``DataFrame``, a two dimensional array for storing heterogeneous data
 
 Although data can be heterogeneous within a ``DataFrame``, the contents of the columns must be homogeneous
+(of the same type)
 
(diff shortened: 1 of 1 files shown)


rst_files/data_statistical_packages.rst
```diff
@@ -51,7 +52,7 @@ The first is to setup columns and construct a dataframe by assigning names
     last_price = [4.2, 11.3, 12.1, missing]
     df = DataFrame(commod = commodities, price = last_price)
 
-Columns of the DataFrame can be accessed by name using a symbol or a ``.``
+Columns of the ``DataFrame`` can be accessed by name using a symbol or a ``.``
```
natashawatkins: is a symbol the colon thing? Have we previously defined this?

arnavs: Yeah, that's a symbol. We define it in introduction to types which comes before

arnavs: Probably good to put in an example, though. Done.

arnavs:
maybe something like this for the overview- 

```
This lecture explores some of the key packages for working with data and doing statistics in Julia

In particular, we will examine the ``DataFrame`` object in detail (i.e., construction, manipulation, querying, visualization, and nuances like missing data) 

While Julia is not an ideal language for pure cookie-cutter statistical analysis, it has many useful packages to provide those tools as part of a more general solution

Examples include ``GLM.jl`` and ``FixedEffectModels.jl``, which we discuss

This list is not exhaustive, and others can be found in organizations such as `JuliaStats <https://github.com/JuliaStats>`_, `JuliaData <https://github.com/JuliaData/>`_, and  `QueryVerse <https://github.com/queryverse>`_
```

arnavs:
Great, merging now

PR #387, opened 2018-11-21
edits to optimization_solver_packages.rst

natashawatkins:
* Could the exercise instead just be a comment at the bottom?


rst_files/optimization_solver_packages.rst
```diff
@@ -444,12 +449,10 @@ For example, if we plot the function
 
 with :math:`x \in [0,1]` we get
```
natashawatkins: Is there meant to be code for a figure here?

arnavs:
Hey @natashawatkins 

1. I agree, the exercise should be a comment, so I made it an "additional notes." We should put in some concrete exercises though on the use of the packages (e.g., "optimize X function using Y package...", or "implement this optimization problem with these constraints in this package...") I'll talk to @jlperla about it.

2. Done.

PR #395, opened 2018-11-22
edits to rst_files/version_control.rst

natashawatkins:
Some of the screenshots are hard to read and probably don't require it to be that zoomed out

### DIFF
--- a/rst_files/version_control.rst
+++ b/rst_files/version_control.rst
@@ -6,6 +6,8 @@
 Git, GitHub, and Version Control
 ******************************************
 
+.. contents:: :depth: 2
+
 Co-authored with Arnav Sood
 
 An essential part of modern software engineering is using version control
@@ -21,7 +23,6 @@ We use version control because
 
 In this lecture, we'll discuss how to use Git and GitHub
 
-.. contents:: :depth: 2
 
 Setup 
 ==================
@@ -40,7 +41,7 @@ Setup
 
     2. Install the `GitHub Desktop <https://desktop.github.com/>`_ application
 
-3. Optionally (but strongly recommended):  On Windows, change the default line-ending by
+3. Optionally (but strongly recommended):  On Windows, change the default line-ending by:
 
     1. Opening a Windows/Powershell console, or the "Git Bash" installed in the previous step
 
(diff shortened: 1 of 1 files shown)


rst_files/version_control.rst
```diff
@@ -84,17 +85,15 @@ Commits
...
+`Here <https://github.com/QuantEcon/lecture-source-jl/commit/ba59c3ea9a0dec10def3f4f3928af5e2827f3b92/>`_
+ is an example of a commit, which revises the style guide in a QuantEcon repo
 
 In particular, commits have the following features
 
 * An ID (formally, an "SHA-1 hash")
 * Content (i.e., a before and after state)
 * Metadata (author, timestamp, commit message, etc.)
 
-It is crucial to remember that commits represent differences in text, as opposed to repository states
```
arnavs: I think we need to keep this, since this is consistently one of the most confusing things about git

rst_files/version_control.rst
```diff
@@ -39,7 +40,7 @@ Setup
 
     2. Install the `GitHub Desktop <https://desktop.github.com/>`_ application
 
-3. Optionally (but strongly recommended):  On Windows, change the default line-ending by
+3. Optionally (but strongly recommended):  On Windows, change the default line-ending by:
```
arnavs: I thought the style was no-colons/minimal punctuation. Don't mind either way, though.

rst_files/version_control.rst
```diff
@@ -237,11 +244,9 @@ Let's unpack the different components
 
 * The *labels* dropdown lets you tag the issue with labels visible from the issues page, such as "high priority" or "feature request" 
 
-* It's possible to tag other issues and collaborators (including in different repos) by linking to them in the comments. This is part of what's called *GitHub-Flavored Markdown* 
+* It's possible to tag other issues and collaborators (including in different repos) by linking to them in the comments -- this is part of what's called *GitHub-Flavored Markdown* 
 
-For an example of an issue, see `example issue <https://github.com/quanteconuser/example_repository/issues/1>`_ 
-
-The checkbox idiom is a common one to manage projects in GitHub
```
arnavs: I think we also need to talk about the checkboxes since it's so common

rst_files/version_control.rst
```diff
@@ -280,7 +285,7 @@ GitHub will give us the following window
 * The warning symbol next to the file indicates the existence of a merge conflict 
 * The viewer tries to show us the discrepancy (I changed the word repository to repo, but someone else tried to change it to "repo" with quotes)
 
-To fix the conflict, we can go into a text editor (such as Atom or VS Code). Here's an image of what we see in Atom 
+To fix the conflict, we can go into a text editor (such as Atom or VS Code)
```
arnavs:
```suggestion
To fix the conflict, we can go into a text editor (such as Atom)
```

Guess I'll just eliminate the mention of VS Code here, give them less options `:-)`

rst_files/version_control.rst
```diff
@@ -384,7 +392,7 @@ You'll see something like this
 .. figure:: /_static/figures/git-create-pr.png
     :scale: 60%
 
-This gives us a quick overview of the commits we want to merge in, as well as the end-to-end differences
+This gives us a quick overview of the commits we want to merge in, as well as the file differences
```
arnavs: I wanted to stress that these differences were "end-to-end," i.e. that they are path-independent/independent of the commit history. Maybe something like "final differences"? Dunno.

rst_files/version_control.rst
natashawatkins: Can you clarify what this means?

arnavs:
Thanks @natashawatkins, think this is solid feedback. See above for some comments. 

I also added the PR image back in when resolving the merge conflict. I'll try to make it bigger/zoom in, but since the PRs have consistently been a pain point for the students, I think it's important to make it as clear as possible with screenshots.

rst_files/version_control.rst
natashawatkins: Ok, I'd profer to just say 'Checkboxes are useful to manage project tasks'

rst_files/version_control.rst
natashawatkins: Sure

rst_files/version_control.rst
natashawatkins: I see what you mean, but not sure if end-to-end really works - maybe overall?

rst_files/version_control.rst
natashawatkins: True - but in this case, I think I found it confusing because there's it's a list in a list... not sure

rst_files/version_control.rst
arnavs: Basically the idea is that a commit is less like a point in time (say `1`), and more like a move between points in time (say `1 --> 2`). So commits aren't versions, but are rather actions.

rst_files/version_control.rst
arnavs: I like overall.

rst_files/version_control.rst
arnavs: That works.

rst_files/version_control.rst
arnavs: Yeah, I have no issue with it. Guess we can keep it for now.

PR #396, opened 2018-11-22
fix displayname for Jupyter kernel used in construction of website

mmcky:
this needs to be the basename `Julia` for the website to properly construct the `breadcrumbs` navigation tool. It will still latch onto `Julia-1.0` kernel.


arnavs:
When I compile the lectures and load them again, it briefly warns "No Kernel!" before defaulting to the Julia 1.0.2. Is this normal?

arnavs:
Poses no serious issue so not worth holding this up

PR #397, opened 2018-11-22
Pooya

PooyaFa:


rst_files/generic_programming.rst
```diff
@@ -426,7 +426,7 @@ Numbers and Algebraic Structures
 
 Define two binary functions,  :math:`+` and :math:`\cdot`, called addition and multiplication -- although the operators can be applied to data structures much more abstract than a ``Real``
 
-In mathematics, a `ring <https://en.wikipedia.org/wiki/ring_\(mathematics\)>`_ is a set with associated additive and multiplicative operators where
+In mathematics, a `ring <https://en.wikipedia.org/wiki/ring_\(mathematics\>`_ is a set with associated additive and multiplicative operators where
```
jlperla: Are you sure that this escapes the links properly?  I am not sure if we can check this yet with sphinx, so I suspect it is premature to modifyu it.

rst_files/generic_programming.rst
```diff
@@ -569,7 +569,7 @@ Hopefully this showcases the power of generic programming:  with a well-designed
 Reals and Algebraic Structures
 =======================================
  
-Thinking back to the mathematical motivation, a `field <https://en.wikipedia.org/wiki/field_\(mathematics\)>`_ is a ``ring`` with a few additional properties, among them
+Thinking back to the mathematical motivation, a `field <https://en.wikipedia.org/wiki/field_\(mathematics\>`_ is a ``ring`` with a few additional properties, among them
```
jlperla: Ditto

rst_files/generic_programming.rst
```diff
@@ -426,7 +426,7 @@ Numbers and Algebraic Structures
 
 Define two binary functions,  :math:`+` and :math:`\cdot`, called addition and multiplication -- although the operators can be applied to data structures much more abstract than a ``Real``
 
-In mathematics, a `ring <https://en.wikipedia.org/wiki/ring_\(mathematics\>`_ is a set with associated additive and multiplicative operators where
+In mathematics, a `ring <https://goo.gl/unRJRB>`_ is a set with associated additive and multiplicative operators where
```
jlperla: I don't really like this approach.  I would rather people be able to copy/see the URL, assuming that the escaping of RST is fixed.  But if there is no other way, then so be it.

rst_files/generic_programming.rst
PooyaFa: @jlperla it doesn't work. I changed it. I will check it now.

jlperla: I don't think that the sphinx/jupinx has been fixed, so checking is premature.

rst_files/generic_programming.rst
PooyaFa: this will work. This is my last change. 

The other formats didn't work properly.

PooyaFa: Ok then it is a different issue for sphinx/jupinx from what we discovered before. Because, external links for websites work all the time, except for web addresses which have "("  ")", I think.

PR #400, opened 2018-11-22
Unit tests for amss

ajozefiak:


arnavs:
Yup, this is exactly what I meant. Thanks for the excellent PR

PR #417, opened 2018-11-26
update default to drop tests and add jupyter-tests

mmcky:
To generate lecture set with tests you will now need to use `jupyter-tests`.


arnavs:
Perfect, thank you.

PR #421, opened 2018-11-27
Amss plots

ajozefiak:


arnavs:
Thanks @ajozefiak. 

Close #421

PR #428, opened 2018-11-27
Fixes

Nosferican:


arnavs:
Thanks. I think at this stage we can commit directly to `master`.

PR #432, opened 2018-11-28
kalman plots

ajozefiak:
The plots have been modified to match the original plots more closely.

Namely: adding back the grey contour lines and changing the colour scheme.

One thing I noticed (that may be an issue) is that the newdist grid messes with the dist grid and the predicteddist grid. Notably, in the third plot I had to bump up the number of contour lines for the dist grid to 14 as opposed to 6, as having 6 contours meant that only 2 or 3 of them were in the dimensions of the figure. For the last figure I omitted the newdist grid since it caused the predictdist surface to lose some of its granularity, going from several colours to only 2 or 3.

### DIFF
--- a/rst_files/kalman.rst
+++ b/rst_files/kalman.rst
@@ -129,9 +129,11 @@ This density :math:`p(x)` is shown below as a contour map, with the center of th
     two_args_to_pdf(dist) = (x, y) -> pdf(dist, [x, y]) # returns a function to be plotted
 
     # plot
-    contour(x_grid, y_grid, two_args_to_pdf(dist), fill = true, levels = 6, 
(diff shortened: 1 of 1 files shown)


ajozefiak:
It's almost as though the newdist grid is scaled in a way it shouldn't be compared to the original lectures. Or, this could just be an issue with gr and plots.jl.

arnavs:
Thanks @ajozefiak. Made a small tweak or two (in the current setup, I think it looks better without the background fill; eventually we should figure out how to set the `fillalpha` and `gridalpha` properly). 

Saw what you mean about the plot settings. Thankfully our regression tests guarantee that the underlying data is valid, so we can tweak things like this without worrying too much.

PR #466, opened 2018-12-03
Update to align Numfocus badge

DrDrij:
Simply adds an id in the HTML for the Numfocus badge allowing CSS styling.


arnavs:
Looks great, thanks @DrDrij

PR #472, opened 2018-12-06
further edits to getting_started.rst

natashawatkins:
I went through it again and simplified the lecture further
* We could probably add more screenshots

### DIFF
--- a/rst_files/getting_started.rst
+++ b/rst_files/getting_started.rst
@@ -29,18 +29,18 @@ Like Python, and unlike Matlab and Stata, there is a looser connection between J
 Because of this, you have much more flexibility in how you write and edit your code, whether
 that be locally or on the cloud, in a text-editor or IDE, etc.
 
-Several interfaces are built upon `Jupyter <http://jupyter.org/>`_  which provides 
-a browser-based interface to access computational kernels for different languages (e.g. Julia, Python, R, etc.)
+One example is `Jupyter <http://jupyter.org/>`_ , which provides 
+a browser-based interface to execute code in different languages (e.g. Julia, Python, R, etc.)
 
 While you will eventually use other editors, there are some advantages to starting with Jupyter while learning Julia
 
-* Nicely formatted output in the browser, including tables, figures, animation, video, etc.
+* The ability to mix formatted text (including mathematical expressions) and code in a single document
 
-* The ability to mix in formatted text and mathematical expressions between cells
+* Nicely formatted output including tables, figures, animation, video, etc.
 
-* Functions to generate PDF slides, static HTML, etc.
+* Conversion tools to generate PDF slides, static HTML, etc.
 
-* It requires no installation if you used a cloud-based solution
+* Can be used in the cloud without requiring installation
 
 Whether you end up using Jupyter as your primary work environment or not, you'll find learning about it an excellent investment
 
@@ -53,9 +53,9 @@ Installing Julia and Dependencies
 .. While using the Docker instance is convenient and error-proof, you may eventually want to install things locally
 
 The easiest approach to using Julia with Jupyter on your desktop is to 
-install `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
+install the latest version of `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
 
-* Install Anaconda by `downloading the binary <https://www.anaconda.com/download/>`_ (3.7 version)
+* Install Anaconda by `downloading the binary <https://www.anaconda.com/download/>`_
 
     * Make sure you click yes to "add Anaconda to my PATH"
 
@@ -64,14 +64,21 @@ install `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
     * We do not recommend `JuliaPro <https://juliacomputing.com/products/juliapro.html>`_
       due to its limited number of available packages
 
-* Open Julia, by navigating to Julia through your menus or desktop icons
+* Open Julia by navigating to Julia through your menus or desktop icons or 
+  typing ``julia`` into a terminal or windows console
 
-Either way you should now be looking at something like this (modulo your operating system)
+Either way you should now be looking at something like this
 
 .. figure:: /_static/figures/julia_term_1.png
    :scale: 100%
 
-This is called the JULIA *REPL* (Read-Evaluate-Print-Loop), which we discuss more :ref:`below <julia_repl>`
+This is called the JULIA *REPL* (Read-Evaluate-Print-Loop)
+
+We examine the REPL and its different modes in more detail in the :doc:`tools and editors <tools_editors>` lecture  
+
+We'll start by installing some key Julia packages needed to run code in the lectures
+
+Typing ``]`` into the REPL switches to the package manager, where we can add packages
 
 * In the Julia terminal, type the following
 
@@ -80,15 +87,7 @@ This is called the JULIA *REPL* (Read-Evaluate-Print-Loop), which we discuss mor
 
         ] add IJulia InstantiateFromURL; precompile
 
-This installs enough Julia packages to begin using the lecture notes
-
 *Note:* On OS/X you will need to type the ``]`` separately and cannot copy/paste the whole string
-
-* To run Jupyter, open a terminal or windows console, ``cd`` to the location you wish to write files and type 
-
-.. code-block:: none
-
-    jupyter lab
 
 
 .. _jl_jupyter:
@@ -102,8 +101,13 @@ Using Jupyter
 Getting Started
 -----------------------
 
-After you have started Jupyter, your web browser should open to a page on the 
-local machine that looks something like this
+To run Jupyter, open a terminal or windows console, ``cd`` to the location you wish to write files and type 
+
+.. code-block:: none
+
+    jupyter lab
+
+Your web browser should open to a page that looks something like this
 
 .. figure:: /_static/figures/starting_nb_julia.png
    :scale: 100%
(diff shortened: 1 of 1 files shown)


rst_files/getting_started.rst
```diff
@@ -53,9 +53,9 @@ Installing Julia and Dependencies
 .. While using the Docker instance is convenient and error-proof, you may eventually want to install things locally
 
 The easiest approach to using Julia with Jupyter on your desktop is to 
-install `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
+install the latest version of `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
 
-* Install Anaconda by `downloading the binary <https://www.anaconda.com/download/>`_ (3.7 version)
+* Install Anaconda by `downloading the binary <https://www.anaconda.com/download/>`_
```
natashawatkins: Easier to say the latest version that keep updating this page

rst_files/getting_started.rst
```diff
@@ -224,41 +214,25 @@ Now try copying the following into a notebook cell and hit ``Shift-Enter``
...
-
-Let's go over some more Jupyter notebook features --- enough so that we can press ahead with programming
-
 
 Tab Completion
 ^^^^^^^^^^^^^^^^^^
 
-A simple but useful feature of `IJulia <https://github.com/JuliaLang/IJulia.jl>`_ 
-(the Jupyter kernel for Julia) is tab completion
+Tab completion in Jupyter makes it easy to find Julia commands and functions available
```
natashawatkins: I'm hesitant to say `IJulia` because it can be confusing for a beginner, the subtle difference between julia, IJulia, and Jupyter. I think it's fine to just say Jupyter because if you're using julia in Jupyter, you're using IJulia.

rst_files/getting_started.rst
```diff
@@ -224,41 +214,25 @@ Now try copying the following into a notebook cell and hit ``Shift-Enter``
...
-IJulia offers up the possible completions
-
-This helps remind you of what's available and saves a bit of typing
+   :scale: 80%
 
 
 .. _gs_help:
 
-Online Help
+Getting Help
```
natashawatkins: Doesn't help also work offline? Also I think we could add a screenshot to this section

rst_files/getting_started.rst
```diff
@@ -270,38 +244,39 @@ Documentation should now appear in the browser
...
 
-In addition to executing code, the Jupyter notebook allows you to embed text, equations, figures and even videos in the page
+In addition to executing code, the Jupyter notebook allows you to embed text, 
+equations, figures and even videos in the document
 
 For example, here we enter a mixture of plain text and LaTeX instead of code
 
 .. figure:: /_static/figures/nb6_julia.png
-   :scale: 100%
+   :scale: 80%
```
natashawatkins: These screenshots should be zoomed out and include the 'In:' 'Out:' flags because they were kind of confusing to interpret

rst_files/getting_started.rst
```diff
@@ -336,38 +303,6 @@ which provides a static HTML representations of notebooks
 
 QuantEcon also hosts the `QuantEcon Notes <http://notes.quantecon.org/>`_ website, where you can upload and share your notebooks with other economists and the QuantEcon community
 
-.. _julia_repl: 
-
-The REPL
```
natashawatkins: I didn't really see the need for this section... it seemed a bit redundant and possibly to much info for getting started

jlperla:
@natashawatkins  A lot of the code in this section was from an older version of the notes.  I left much of it in because I thought it better than stripping things out entirely (i.e., there may have been good reasons for writing it in the first place that I didn't want to second-guess).

So in principle, other than the basic setup and notebook tutorial, I am open-minded to moving content.

natashawatkins:
Yeah I think the shorter the better for this lecture

rst_files/getting_started.rst
jlperla: Normally I prefer to strip things out of the introduction.  The problem is that without the REPL section, I think people will be very confused if they need to do package operations or people read about the REPL online.

One option is to tag it as `(Advanced)` in the title i.e. `The REPL (Advanced)` or something like that.   Then readers know it is somewhat optional.

Otherwise, there is no where else they would see the REPL unless they read the (optional) advanced julia section.

jlperla: But I think what is in there could be simplified a bit bit.

natashawatkins: I thought the REPL was mentioned earlier when you open Julia and switch the package manager

I also added earlier that you can open julia by typing `julia` in the REPL

rst_files/getting_started.rst
```diff
@@ -206,16 +176,36 @@ If it's a long file, however, you have the alternative of
 The present working directory can be found by executing the command ``pwd()``
 
 
+Using QuantEcon Lecture Packages
+-------------------------------------------
+
+To use the curated set of packages in the QuantEcon lecture notes, 
+copy the following text into a notebook cell, and hit ``Shift-Enter`` to run the cell
+
+   .. literalinclude:: /_static/includes/deps.jl
```
jlperla: @arnavs I would prefer to move this to the dependencies without includes literalinclude.  Arnav, can you help her out with that?

natashawatkins: We had already run this code earlier in the lecture I think...?

rst_files/getting_started.rst
```diff
@@ -310,14 +285,6 @@ contents of the current working directory
 These shell commands are handled by your default system shell and hence are platform specific
 
 
-Package Manager
-^^^^^^^^^^^^^^^^
-
-You can enter the package manager by prepending a ``]``
-
-For example, ``] st`` will give the status of installed packages in the current environment
```
jlperla: I don't like ripping this out.  People need to know about the package manager, and there is nowhere else they would learn it.

natashawatkins: Earlier in the lecture we've basically covered all this info when we initially add the Instantiate package

rst_files/getting_started.rst
```diff
@@ -336,38 +303,6 @@ which provides a static HTML representations of notebooks
...
-This is a Julia specific terminal disconnected from the graphical interface of 
-Jupyter, and becomes increasingly important as you learn Julia
-
-For example, the REPL is what we used in the beginning of this lecture to install 
-``InstantiateFromURL.jl`` and ``IJulia``
-
-To start the REPL in a typical Jupyter lab environment
-
-#. Choose "New Launcher"
-#. Choose a ``Julia 1.0`` Console
```
jlperla: This stuff is not obvious to people, and is sometimes necessary.  Again, the issue is that all notebook and REPL based stuff has been moved out of the getting started otherwise.

But perhaps it could be condensed a little.

rst_files/getting_started.rst
```diff
@@ -336,38 +303,6 @@ which provides a static HTML representations of notebooks
...
-Otherwise, if you have a local installation, then  
-
-#. Navigate to Julia through your menus or desktop icons (Windows, OSX), or
-
-#. Open a terminal and type ``julia`` (Linux)
-
-The REPL is one of the best places to add and remove packages, so a good test is to see the current status of the package manager
-
-.. code-block:: julia
-
```
jlperla: Again, this whole section is necessary for people to know to use julia.  It could all be tagged `Advanced`, but there is currently no other lecture it could go in.

rst_files/getting_started.rst
```diff
@@ -64,14 +64,21 @@ install `Anaconda <https://www.anaconda.com/download/#macos>`_ and then Julia
...
    :scale: 100%
 
-This is called the JULIA *REPL* (Read-Evaluate-Print-Loop), which we discuss more :ref:`below <julia_repl>`
+This is called the JULIA *REPL* (Read-Evaluate-Print-Loop)
+
+We examine the REPL and its different modes in more detail in the :doc:`tools and editors <tools_editors>` lecture  
+
+We'll start by installing some key Julia packages needed to run code in the lectures
+
+Typing ``]`` into the REPL switches to the package manager, where we can add packages
```
natashawatkins: @jlperla here is where the REPL is mentioned - I personally think this sufficient

rst_files/getting_started.rst
jlperla: I think this is a good time to go through all of the screenshots to see what requires a zoom-in/etc.

rst_files/getting_started.rst
natashawatkins: Here is where we already have run the dependencies

rst_files/getting_started.rst
jlperla: It was used in the setup instructions, but was a mechanical step that they may have thought was a one-time thing to do.  It wasn't described as something that they would use otherwise, nor was it explained how to access it from a jupyterhub installation.  In particular, we want them to associate package operations with using the REPL

As I said, though, it could potentially be condensed.

rst_files/getting_started.rst
jlperla: I think you need to be very careful walking through the instructions berfore making radical changes.  Those instructions were done as part of the setup routine, not in a notebook.  These were all tested carefully, so we want to make sure we don't break the workflow.

rst_files/getting_started.rst
jlperla: That was a one-time setup.

rst_files/getting_started.rst
natashawatkins: I think we should explain in the setup - it can be discouraging to just be entering commands without knowing what's going on, so I think we should describe the REPL and package manager when they're needed

rst_files/getting_started.rst
jlperla: It ended up being frequently necessary in the course we taught.  There is no other place in the lectures for this sort of stuff since we tell people that the "Tools and Editors" is optional.  But the section on the REPL can be marked `advanced` so that people can refer back to it when they need it.

One option is to tell people how to launch the REPL and then point them at the https://lectures.quantecon.org/jl/tools_editors.html# sections for doing a few operations.  But I am hesitant to say we have given people the tools to use julia without telling them that they can (and sometimes should) play in the REPL.  This section has already been significantly reduced from the older version of the notes.

rst_files/getting_started.rst
jlperla: As I said, I am very confused right now with your suggested changes.  Can you restart your PR by putting in the many excellent edits which don't have to do with flow changes in the desktop setup vs. jupyter setup steps?

Then we can see what you are proposing for the reorganization of the other material in isolation.  I may agree with a lot of it now that I have hopefully made the point that the setup and usage sections need to be distinct.

rst_files/getting_started.rst
jstac: Regarding the REPL section, we don't need to tell people how to open the REPL, given that we already assume they can open it at the top of the lecture.  I suggest cutting those instructions.

On the topic of making this lecture more friendly, I recommend a few more sentences after ``] add IJulia InstantiateFromURL; precompile`` 

I hate typing something into my fresh clean installation when I don't know what's going on...

Just a couple of dot points would be great, saying that

1. `]` activates the package manager 
2. Julia needs to install additional code in order to run the programs in the lectures
3. This code comes in libraries called packages, about which we will say more below
4. The QE team has organized a collection of useful packages, and these instructions install them in one go
5. You can find the list of packages installed by this command by typing ??

jlperla: The problem is that many people (i.e. any students I have or the schools we suggest setup jupyterhub) will not actually run through that setup process.  It is only for the desktop environment and not docker or jupyterhub.  Launching the REPL has different steps in a jupyterhub installation.

We can put a few more sentences if you like.  I have the opposite preference for setup instructions (i.e., I like to be told exactly what to do with little exposition, and blindly follow instructions) but you have a defensible position.  I don't think it causes much damage here either way.

For your text, another place your text could work is when we use the `InstantiateFromURL` package (which happens for desktop, jupyterhub, etc.).  At that point in the instructions, the `add IJulia` is basically just to get Jupyter up and running in a minimal setup step, and the InstantiateFromURL is used to get the packages.  Having the extra discussion in the setup might be confusing since the whole `IJulia` step is not quantecon specific.

Another approach is to have the `] add InstantiateFromURL` from within Jupyter so that the setup step prior to jupyter is just `] add IJulia; precompile`.  This may be an easier flow, doesn't break the jupyterhub setups, and your text makes a lot of sense.

rst_files/getting_started.rst
natashawatkins: @arnavs could you help with this?

arnavs: Hey, yeah this definitely works offline.

arnavs: Done. I'm also glad you brought this up, because `repmat` has been deprecated in favor of `repeat`, so people would be especially confused...

arnavs:
Closing this because it looks like we've included this content in the recent refactor.

PR #475, opened 2018-12-07
remove references to IJulia in getting_started.rst

natashawatkins:
As discussed in #472


arnavs:
Yup, no need to mention IJulia

PR #477, opened 2018-12-07
WIP: Discussion about reorganizing setup and REPL workflow

jlperla:
cc: @jstac @arnavs 

@natashawatkins  This is a PR to discuss issues on the reorganization with the REPL and setup instructions.  After we discuss things from the perspective of different user scenarios, we can make changes to the flow/structure.  Basically, lets discuss and implement stuff from #472

To give some background: this lecture should really be called "Introduction to the Julia Environment".  The original lecture had way more content, and I tried to move a good chunk of it over to the new https://lectures.quantecon.org/jl/tools_editors.html# section which I didn't think was necessary for people to use the notes and do julia basics.  There may be things that I could have moved over but didn't.

But there are a few constraints here:  first, the `tools_editors` lecture is completely optional, and I suspect that most people will not read it until much later in a course.  There are no other lectures that have julia environment examples, so if there is something that we think people really need to know, then either (1) we create a new mini-lecture in the "intro to julia" section, or (2) we keep it in here.

The next one is that I think many students (all if I had my way!) will not be doing a Julia desktop installation.  This means we cannot assume they will follow the setup instructions in https://lectures.quantecon.org/jl/getting_started.html#Installing-Julia-and-Dependencies  

For most students on jupyterhub, they would be pointed directly to https://lectures.quantecon.org/jl/getting_started.html#Using-Jupyter 

Finally, we need to keep in mind that while we hope students are using Jupyter almost exclusively, as soon as they start to do web searches on Julia they will hear about the package manager, the REPL, etc.   Furthermore, the instructor/TAs will almost certainly have a REPL they are demonstrating with in class and doing package operations/etc.  This is very confusing to them if they have no exposure to the REPL and package manager.

Let me post specific discussions in separate comments.

### DIFF
--- a/rst_files/getting_started.rst
+++ b/rst_files/getting_started.rst
@@ -338,8 +338,8 @@ QuantEcon also hosts the `QuantEcon Notes <http://notes.quantecon.org/>`_ websit
 
 .. _julia_repl: 
 
-The REPL
-------------
+The REPL (Advanced)
+---------------------
 
 While we have not emphasized it, on any :ref:`JupyterHub <jl_jupyterhub>` or local Jupyter 
 installation you will also have access to the Julia REPL


jlperla:
The desktop setup instructions currently have you load up the REPL.  The only purpose of this is to install IJulia so you can use jupyter.  Otherwise, we would just have them jump to jupyter directly and skip the REPL sub-step.

Right now it is
```julia
] add IJulia InstantiateFromURL; precompile
```
which does both the `InstantiateFromURL` required for using the lecture notes, and also jupyter support.  I think that @jstac  comments in https://github.com/QuantEcon/lecture-source-jl/pull/472#discussion_r239341373 are very helpful, but I am not sure if they belong here or below, since they give a lot of detail on using the package manager for what is a sub-step in the installation.

The other thing is that those comments are sufficiently useful that they should be read by everyone rather than just those doing a desktop installation (i.e. being using jupyterhub should still have that information).

Because of this, my gut is that we should do the following.

Change this step to 
```julia
] add IJulia; precompile
```
with maybe another sentence or two, but minimal, and then add in a

```julia
] add InstantiateFromURL
```
In the Jupyter section, which everyone can read.  And then below that point, we can add in @jstac comments.

Keep in mind that the reason we didn't have that `] add InstantiateFromURL` for everyone was that it is not necessary for a jupyterhub installation... @arnavs  do you think that will mess up a jupyterhub installation?  i.e. if there is a precompiled InstantiateFromURL would that package operation break anything?  If not, then this flow could work well.

jlperla:
Next comment from your PR:  Do we need a section on the REPL?  This goes back to there not being any other lecture in which to have it.  Right now there is very little content on it, other than telling people how to run it in jupyterhub vs. the local desktop.

In this PR I put `(Advanced)` in that title which may help people to realize it is less essential.  Other things we can do:
- Remove the `] st` as the example to use in the REPL.  I don't have a strong feeling, but I think pointing out to people that they probably should do package operations in it is useful.
- Learning to use the REPL for package operations is necessary for people when they start writing their own code, since they need to `] add` packages even after the precompilation step with `] add InstantiateFromURL`.  The reason is that the installed packages are not in their global registry.

jlperla:
For other content to consider removing: The section https://lectures.quantecon.org/jl/getting_started.html#Working-with-Files was from the older notes.  I think we can drop it, but I defer to other people.

We can drop the https://lectures.quantecon.org/jl/getting_started.html#Sharing-Notebooks section as well, but I think it is useful as a plug for QuantEcon Notes and to tell people about `nbviewer`.

arnavs:
Going to dive through this later, but with regards to 

> Keep in mind that the reason we didn't have that `] add InstantiateFromURL` for everyone was that it is not necessary for a jupyterhub installation... @arnavs do you think that will mess up a jupyterhub installation? i.e. if there is a precompiled InstantiateFromURL would that package operation break anything? If not, then this flow could work well.

I don't think this should be OK. The only risk in principle is that we've pushed hotfixes to the package that we haven't tagged yet. But we haven't pushed new code to this package since the tagged release `v0.1.0`

jlperla:
@jstac @natashawatkins 

I played around with a reordering to move the installation of `InstantiateFromURL` to the notebook section.  I did not think it was an improvement, in particular since the first part of the lecture is a one-time setup, `] add InstantiateFromURL` is something you typically would want to do in your initial setup - and only on a desktop installation. 

Since you guys had concerns about the organization, I think we will need to discuss alternatives with you before going any further.  I would rather do that sooner than later, if possible, so we can close out the issue.

One radical option is to split the lecture into two:  (1) minimalist julia setup instructions and (2) an introduction to the julia environment.  But, frankly, I don't think that is an improvement either.  This is not a long lecture to be begin with, and these are typically things you would do all at once.

jstac:
@jlperla I'll try to get back to you on this today.

jlperla:
The way the lecture is setup right now, the jupyter section is setup to be usable from either a desktop or a jupyterhub installation.  The early section is desktop only, so not everyone would step through the steps.

If you don't like this general organization then there are things we can discuss, but merging the two will be very confusing for people coming from different types of installations

jlperla:
@arnavs  is working on this reorganization today.  I think it will take us back towards the original organization with different ways of accessing julia all in one place in the first setup, which is my preference (especially since it will help us steer people towards whatever are the best tested jupyterhub setups at any point in time).

This will also allow us to do an extra step in the desktop setup instructions to resolve a major confusion of users with julia package environments.

arnavs:
Closing this because it's now reflected in the refactor `intro-reorg` PR, which includes content from this and the individual PRs that @jstac mentioned.

PR #492, opened 2018-12-17
fix compilation error in generic_programming.rst

natashawatkins:


arnavs:
Looks good to me, thanks for this @natashawatkins

PR #493, opened 2018-12-17
add discussion of package manager to getting_started.rst

natashawatkins:


jlperla:
The only problem is that this text may not be read by people coming with jupyterhub, who should be skipping this section. I think it is better to put the comments below where `using InstantiateFromURL`, which everyone needs to do. 

The problem blem

PR #494, opened 2018-12-17
move open jupyter in getting_started.rst

natashawatkins:


natashawatkins:
Where do we say for JupyterHub people to skip that section?

jlperla:
Because that jupyter section applies to people doing both a desktop installation and coming in directly from a jupyterhub installation, but those instructions only work for a desktop installation and would confuse any jupyterhub students. Hence why they were in the desktop setup.

Arnav and I will try to reorganize this into two lectures and see if that makes things easier to read.

PR #495, opened 2018-12-17
edits to REPL discussion in getting_started.rst

natashawatkins:


jstac:
Presently we explain how to open the REPL late in the lecture, after already requiring that the user opens the REPL --- which is illogical.

This PR is a fix.

jlperla:
The problem is two fold:
* not everyone goes through the first part of that lecture, and they may be starting from an existing Julia/jupyter installation.
* if you are on jupyterhub, there is a different way to start the REPL. hence the different description

jstac:
> not everyone goes through the first part of that lecture

Let's go ahead and assume people read things in order.  It might not be true in every instance but it's generally regarded as a plausible assumption.

jlperla:
This is largely innocuous since it doesn't change the setup instructions, but I really like the comments. If it was in the lower section then everyone would read it (even if we tell them to skip the setup steps in the first part). I am concerned that even if they read the comments, they may assume it is only relevant for desktop setups

arnavs:
I believe these changes are incorporated in the `intro-reorg` PR. 

That is, in the `getting_started` lecture we have the stuff Natasha moved there (which we need there for package operation). And in the subsequent `julia_environment` lecture we remove references to the desktop setup and say: 

```
The REPL
------------

The desktop-based REPL we saw :ref:`previously <intro_repl>` is also available inside the Jupyter browser

#. Choose "New Launcher"
#. Choose a ``Julia 1.0`` Console
```

@jstac @jlperla @natashawatkins Let me know if you think that makes sense, in which case I'll close this.

arnavs:
Actually, we can discuss this on the PR, no need to have a separate issue. And I think these two changes are compatible (the only difference is that we left the JupyterHub REPL instructions in the REPL section of the Jupyter lecture, instead of deleting them).

PR #506, opened 2018-12-22
replaced ''bellman_operator'' with ''T''.

PooyaFa:
replaced ''bellman_operator'' with ''T''.

### DIFF
--- a/rst_files/ifp.rst
+++ b/rst_files/ifp.rst
@@ -360,7 +360,7 @@ Implementation
 
 Here's the code for a named-tuple constructor called ``ConsumerProblem`` that stores primitives, as well as
 
-* a ``bellman_operator`` function, which implements the Bellman operator :math:`T` specified above
+* a ``T`` function, which implements the Bellman operator :math:`T` specified above
 
 * a ``coleman_operator`` function, which implements the Coleman operator :math:`K` specified above
 
@@ -403,7 +403,7 @@ Setup
         return (r = r, R = R, β = β, b = b, Π = Π, z_vals = z_vals, asset_grid = asset_grid)
     end
 
-    function bellman_operator!(cp, V, out; ret_policy = false)
+    function T!(cp, V, out; ret_policy = false)
 
         # unpack input, set up arrays
         @unpack R, Π, β, b, asset_grid, z_vals = cp
@@ -436,8 +436,8 @@ Setup
         out
     end
 
-    bellman_operator(cp, V; ret_policy = false) =
-        bellman_operator!(cp, V, similar(V); ret_policy = ret_policy)
+    T(cp, V; ret_policy = false) =
+        T!(cp, V, similar(V); ret_policy = ret_policy)
 
     get_greedy!(cp, V, out) =
         update_bellman!(cp, V, out, ret_policy = true)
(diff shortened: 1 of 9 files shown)


PooyaFa:
related to #236

arnavs:
Thanks @PooyaFa for the hard work. I’ll take a look at this tonight or tomorrow.

Don’t forget to enjoy the holidays:)

PooyaFa:
> Thanks @PooyaFa for the hard work. I’ll take a look at this tonight or tomorrow.
> 
> Don’t forget to enjoy the holidays:)

Thank you very much @arnavs . Happy Holidays :)

rst_files/lln_clt.rst
```diff
@@ -214,26 +214,26 @@ Setup
...
     gr(fmt = :png, size = (900, 500))
 
 .. code-block:: julia
 
     function ksl(distribution, n = 100)
         title = nameof(typeof(distribution))
-        observations = rand(distribution, n)
-        sample_means = cumsum(observations) ./ (1:n)
+        X_i = rand(distribution, n)
+        X̄_n = cumsum(X_i) ./ (1:n)
```
arnavs: We need to be careful with this --- I think we ruled against unicode glyphs (combinations of symbols, like letters and bars/hats/underscores) in the style guide.

rst_files/orth_proj.rst
```diff
@@ -70,7 +70,7 @@ For a linear subspace  :math:`S \subset \mathbb{R}^n`, we call :math:`x \in \mat
 .. figure:: /_static/figures/orth_proj_def2.png
     :scale: 50%
 
-The **orthogonal complement** of linear subspace :math:`S \subset RR^n` is the set :math:`S^{\perp} := \{x \in \mathbb{R}^n \,:\, x \perp S\}`
+The **orthogonal complement** of linear subspace :math:`S \subset R^n` is the set :math:`S^{\perp} := \{x \in \mathbb{R}^n \,:\, x \perp S\}`
```
arnavs: Good catch. Although I think should be changed to `\RR^n`? Whatever the rest of them use in this lecture.

rst_files/ifp.rst
```diff
@@ -493,7 +493,7 @@ Setup
         return V, c
     end
 
-Both ``bellman_operator`` and ``coleman_operator`` use linear interpolation along the asset grid to approximate the value and consumption functions
+Both ``T`` and ``coleman_operator`` use linear interpolation along the asset grid to approximate the value and consumption functions
```
arnavs: We probably shouldn't get rid of `bellman_operator` and leave `coleman_operator`. In the lecture we use _K_ to denote that operator (see the `Col90` ref); could you make that shift?

rst_files/orth_proj.rst
PooyaFa: Thanks. Apparently \RR^n isn't work properly. I will change them to \mathbb{R}^n like above. Is it ok?

rst_files/ifp.rst
PooyaFa: Of course. First, I tried to change only bellman_operator, and now I am going through each lecture notes by the order in http://compare-lectures.quantecon.org/jl/index.html.

rst_files/orth_proj.rst
arnavs: Yep, that's perfect. A while back we eliminated the shortcodes (`\RR`, `\EE`, etc.) to just use the raw LaTeX.

rst_files/ifp.rst
arnavs: Gotcha. There's maybe something to be said for doing this per-lecture (otherwise, how do you remember which issues remain in which lecture?) But I'll let you handle the workflow.

arnavs:
Closing this in favor of the other PR

PR #512, opened 2019-01-03
fix broken link

fratrik:
Small broken link to pdf.


arnavs:
Thanks, looks good to me.

Issue #448, opened 2018-11-29
Rearrange lake model to get rid of "function LakeModel" and use named tuple


jlperla:
For consistency, and to establish the pattern for this.  Consider merging of named tuples, for example, or passing both to functions.

PR #515, opened 2019-01-04
revise lakemodel

arnavs:
Cleanup per https://github.com/QuantEcon/lecture-source-jl/issues/448.

### DIFF
--- a/rst_files/lake_model.rst
+++ b/rst_files/lake_model.rst
@@ -267,7 +267,7 @@ Let's observe these matrices for the baseline model
 
 .. code-block:: julia
 
-    A_matrices(lm).A_hat
+    matrices.A_hat
 
 And a revised model
 
--- a/rst_files/lake_model.rst
+++ b/rst_files/lake_model.rst
@@ -217,60 +217,76 @@ Setup
 
 .. code-block:: julia
 
-    function LakeModel(;λ = 0.283, α = 0.013, b = 0.0124, d = 0.00822)
+    LakeModel = @with_kw (λ = 0.283, α = 0.013, b = 0.0124, d = 0.00822)
+
+    function A_matrices(lm)
+        @unpack λ, α, b, d = lm
         g = b - d
-        A = [(1 - λ) * (1 - d) + b (1 - d) * α + b   # carry out intermediate calculations
-            (1 - d) * λ            (1 - d) * (1 - α)]
+        A = [(1 - λ) * (1 - d) + b      (1 - d) * α + b
+            (1 - d) * λ                 (1 - d) * (1 - α)]
         A_hat = A ./ (1 + g)
-        return (λ = λ, α = α, b = b, d = d, g = g, A = A, A_hat = A_hat) # return a NamedTuple
+        return (g = g, A = A, A_hat = A_hat)
     end
 
     function rate_steady_state(lm)
-        sol = fixedpoint(x -> lm.A_hat * x, fill(0.5, 2))
+        A_hat = A_matrices(lm).A_hat
+        sol = fixedpoint(x -> A_hat * x, fill(0.5, 2))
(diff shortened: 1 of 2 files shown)


rst_files/lake_model.rst
```diff
@@ -217,60 +217,76 @@ Setup
...
         return x_path
     end
 
+Let's observe these matrices for the baseline model
+
 .. code-block:: julia
 
     lm = LakeModel()
-    lm.α
+    matrices = A_matrices(lm)
```
jlperla: Having a named tuple called "matrices" is kind of weird and a step away from the math.

Why not just splat the results directly into variables and use them directly?

rst_files/lake_model.rst
```diff
@@ -217,60 +217,76 @@ Setup
...
         for t in 1:T
             X_path[:, t] = X
-            X = lm.A * X
+            X = A * X
         end
         return X_path
     end
 
     function simulate_rate_path(lm, x0, T)
+        A_hat = A_matrices(lm).A_hat
```
jlperla:
```
@unpack A_hat = A_matrices(lm)
```
for example....

rst_files/lake_model.rst
```diff
@@ -217,60 +217,76 @@ Setup
 
 .. code-block:: julia
 
-    function LakeModel(;λ = 0.283, α = 0.013, b = 0.0124, d = 0.00822)
+    LakeModel = @with_kw (λ = 0.283, α = 0.013, b = 0.0124, d = 0.00822)
+
+    function A_matrices(lm)
```
jlperla: I don't like the name.  

Maybe `transition_matrices` instead

rst_files/lake_model.rst
```diff
@@ -217,60 +217,76 @@ Setup
...
+        @unpack λ, α, b, d = lm
         g = b - d
-        A = [(1 - λ) * (1 - d) + b (1 - d) * α + b   # carry out intermediate calculations
-            (1 - d) * λ            (1 - d) * (1 - α)]
+        A = [(1 - λ) * (1 - d) + b      (1 - d) * α + b
+            (1 - d) * λ                 (1 - d) * (1 - α)]
```
jlperla: Need equation reference comment here and everywhere.

rst_files/lake_model.rst
```diff
@@ -217,60 +217,76 @@ Setup
...
         A_hat = A ./ (1 + g)
```
jlperla: Change `A_hat` to `Â` throughout.

This is a general comment.  For any annotations of ascii letters (e.g. bar, hat, etc.) use the appropriate glyph.  Sadly it doesn't work with annotations on greek letters/etc.

rst_files/lake_model.rst
arnavs: Done. I'll also add a note on this to the style guide (i.e., try not to create "fake" objects that don't correspond to something in the math or economics).

rst_files/lake_model.rst
arnavs: Done. I just felt weird using `@unpack` for one variable.

rst_files/lake_model.rst
arnavs: Done.

rst_files/lake_model.rst
arnavs: Done.

rst_files/lake_model.rst
arnavs: Sounds good. So unless you have other concerns, I'll merge this (might as well get this fix up) and create a separate issue for the equation numbering (?). If the thing above is too much work, we can always come back to this.

arnavs:
Merging as it incorporates feedback. Will wait to hear from @mmcky about eqn #s.

Issue #439, opened 2018-11-29
Decrease some output in discrete_dp and convert to @with_kw


jlperla:
In http://compare-lectures.quantecon.org/jl/discrete_dp.html#
- [x] Instead of a `SimpleOG` function, it should be a named tuple generator and a function to generate Q and R from the named tuple generator 
- [x] The output of `SimpleOG()`, etc. should be prevented from display with `;` or the lines should be merged together.
- [x] Didn't I have a verbose and non-verbose one in my version?

PR #516, opened 2019-01-05
revise discrete dp

arnavs:
Clean up per https://github.com/QuantEcon/lecture-source-jl/issues/439.

### DIFF
--- a/rst_files/discrete_dp.rst
+++ b/rst_files/discrete_dp.rst
@@ -473,13 +473,15 @@ Setup
 
 .. code-block:: julia
 
-    using BenchmarkTools, Plots, QuantEcon
+    using BenchmarkTools, Plots, QuantEcon, Parameters
     gr(fmt = :png);
 
 .. code-block:: julia
 
-    function SimpleOG(;B = 10, M = 5, α = 0.5, β = 0.9)
+    SimpleOG = @with_kw (B = 10, M = 5, α = 0.5, β = 0.9)
 
+    function QR_matrices(g)
+        @unpack B, M, α, β = g
         u(c) = c^α
         n = B + M + 1
         m = M + 1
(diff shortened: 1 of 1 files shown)


rst_files/discrete_dp.rst
```diff
@@ -597,9 +642,10 @@ What happens if the agent is more patient?
 
 .. code-block:: julia
 
-    g_2 = SimpleOG(β=0.99)
+    g_2 = SimpleOG(β=0.99);
+    matrices_2 = QR_matrices(g_2);
```
jlperla: I don't like this name.  Call it what it is, not the data structure.  `transition_matrices(...)` for example.
Also, I would prefer either splatting directly into the Q and R matrices, or forgetting the named tuple and just returning multiple arguments.

Notice
```
julia> using Parameters

julia> nt = (a = 2, b = 2)
(a = 2, b = 2)

julia> @unpack a, b = nt
(a = 2, b = 2)

julia> a
2

julia> b
2

julia> f() = (a = 2, b = 2)
f (generic function with 1 method)

julia> a, b = f()
(a = 2, b = 2)

julia> a
2

julia> b
2
```
i.e. you can return a named tuple into multiple arguments without an unpack.  But if you are only using a simple function once or twice, you can just return the tuple.

rst_files/discrete_dp.rst
```diff
@@ -597,9 +642,10 @@ What happens if the agent is more patient?
...
 
-    ddp_2 = DiscreteDP(g_2.R, g_2.Q, g_2.β)
+    ddp_2 = DiscreteDP(matrices_2.R, matrices_2.Q, g_2.β)
```
jlperla: after the above, just use R and Q directly.

arnavs:
OK @jlperla see now.

arnavs:
Merging as Jesse's feedback is now incorporated.

Issue #487, opened 2018-12-14
Simplify docker setup


jlperla:
This has too many steps: https://lectures.quantecon.org/jl/tools_editors.html#Docker

I don't think there is any value in having a "no frills" approach.  Just tell people what you want them to do with the minimal number of steps to do it.

arnavs:
OK, will do.

PR #517, opened 2019-01-05
streamline quantecon/base setup instructions

arnavs:
Clean up per https://github.com/QuantEcon/lecture-source-jl/issues/487.


arnavs:
Merging as this doesn't change any workflow steps and all these instructions have already been tested.

PR #519, opened 2019-01-07
rewrite uncertainty traps

arnavs:
Eliminate the `mutable struct` and push to a dataframe instead 

Also, move the functions like `gen_aggregates` and `new_beliefs` inside the `simulate` closure

### DIFF
--- a/rst_files/uncertainty_traps.rst
+++ b/rst_files/uncertainty_traps.rst
@@ -405,22 +405,8 @@ is, the number of active firms and average output
         # draw standard normal shocks
         w_shocks = randn(capT)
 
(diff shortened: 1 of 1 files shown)


arnavs:
@Nosferican @jlperla Requested your review on this. The key change was moving from the `mutable struct` to pushing to a dataframe and the closure above. Let me know if this looks OK.

arnavs:
Made some simplifications per @jlperla. We'll have to think more carefully about how to simplify this further. Per Slack, will merge this in now (as it's a strict improvement), and we can add it to the list of cleanups.

PR #529, opened 2019-01-17
Equality in Tests 

XiaojunGuan:


arnavs:
Thanks @XiaojunGuan. I made a small tweak based on a failing test (can't use approximation with tuples). But otherwise looks great; this will be a great help for the regression testing.

And thanks for cleaning up some of the comments as well.

PR #535, opened 2019-01-29
Statsplots shift WIP

arnavs:
Whitespace-free diff is [here](https://github.com/QuantEcon/lecture-source-jl/pull/535/files?utf8=%E2%9C%93&diff=unified&w=1).

### DIFF
--- a/rst_files/generic_programming.rst
+++ b/rst_files/generic_programming.rst
@@ -57,11 +57,11 @@ Setup
 
 .. literalinclude:: /_static/includes/deps_no_using.jl
 
-.. code-block:: julia 
-    :class: hide-output 
+.. code-block:: julia
+    :class: hide-output
 
-    using LinearAlgebra, Statistics, Compat 
-    using Distributions, StatPlots, QuadGK, Polynomials, Interpolations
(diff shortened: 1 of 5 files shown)


arnavs:
@XiaojunGuan, could you run regression tests on these lectures? 

Basically, we need to: 

(1) Change the `activate_github` statements at the top to eliminate the `tag = ""` argument (so we use `master`, which has `StatsPlots` instead of `StatPlots`). 

(2) Run the files changed, confirming that (a) the tests pass and plots show up (so build with `make jupyter-tests`), and (b) we don't see any `precompiling` output that isn't on the website. 

Let me know if stuff is unclear.

XiaojunGuan:
screen shot 2019-02-05 at 3 52 44 pm
screen shot 2019-02-05 at 3 52 53 pm
screen shot 2019-02-05 at 3 53 00 pm
screen shot 2019-02-05 at 3 53 09 pm

arnavs:
Thanks @XiaojunGuan. The precompiling output here is a result of packages not being installed on your machine, so we're OK. 

Merging as the tests pass and the plots look good.

PR #548, opened 2019-03-11
Parallelization [WIP]

arnavs:


arnavs:
@ajozefiak One useful feature would be a way to budget the number of cores or CPU %, or something like that. Is this possible? Otherwise it sucks up all the computing power my computer has. 

But this is a great first step, it's definitely working in parallel. We'd need to see that nothing is screwy on the precompilation side. But if this checks out, we could add it into our Jupinx docker images for parallel generation/execution.

ajozefiak:
Yes, it can be parallelized such that a specific number of cores is utilized. In terms of the pre-compilation side, my intuition says it should be fine, since every notebook is running as a separate process and so the OS should take care of everything there (though I'm not sure about Docker). Maybe the only awkward thing is that messages coming back from the individual processes are asynchronous so the order is somewhat lost.

arnavs:
No worries about order. For safety, I'd like to put in a precompile/activate step at the top, but that's trivial. 

Once you get in the compute budget, I'll test it and hopefully we can merge. 

Thanks for the hard work.

ajozefiak:
@arnavs I've pushed a new commit where the script uses semaphores in order to create at most N concurrent processes.

arnavs:
Thanks @ajozefiak, I will review tonight and then hopefully merge.

arnavs:
Perfect, can confirm the budgeting logic works. Thanks for doing this; now we can execute in parallel without going through the whole build repo. 

cc @jlperla

PR #551, opened 2019-03-27
Template badges

DrDrij:
This PR removes the "How to read this lecture..." and lecture badges from this source repo and moves them to the nbconvert build template. The reasoning being those sections are only needed in the html version of lectures and should not be written to the ipynb files.

Added an optional include file header.raw which will display a QE logo in the top right (first cell) of lecture notebooks.


arnavs:
Thanks @DrDrij, I think @jlperla will appreciate locally-built notebooks without those headers. I tried this out on my machine and it looks much cleaner.

I deleted the now-empty `lecture_howto_jl_header.raw`, because your file captures it.

PR #554, opened 2019-05-15
Download config

AakashGfude:
config variables to allow/disallow download and specifying download path


mmcky:
@arnavs this is being done as a change in the extension. We are now building a separate `downloads` set of jupyter notebooks as the prepend url path to make them work (referring to the website) and not disrupt the test `preview` server by forcing links to the live site rather than local relative links. It will make previewing the html locally a lot easier.

arnavs:
@mmcky Sounds good to me. Merging now.

PR #555, opened 2019-05-17
Rewrite Dynamic Stackelberg 

### LINKED ISSUE #544: Rewrite Dyn_Stack
Per John, update the lecture to pull in Tom's rewrite on the [Python side](https://lectures.quantecon.org/py/dyn_stack.html). 

Probably a good way to break this up is: 

- [ ] Minimal port (tests pass, code is fairly "Julian," text and math and equation references are up to date). 

- [ ] Take a second look at the Julia code for performance/clarity.

Will assign to myself for now.

arnavs:
Rewrites the lecture to match Tom's edits on the Python side. Would resolve #544  

All the plots and numbers check out against the Python, and I added tests here and there where appropriate. 

Once someone gives this the OK, we can merge it in. 

Edit: When building, you can run `make jupyter-tests` to see the notebook with the tests we have in. But those aren't visible to end users.

cc @jstac

### DIFF
--- a/rst_files/dynamic_programming_squared/dyn_stack.rst
+++ b/rst_files/dynamic_programming_squared/dyn_stack.rst
@@ -8,97 +8,310 @@ Dynamic Stackelberg Problems
 
 .. contents:: :depth: 2
 
-Overview
-========
+This notebook formulates and computes a plan that a **Stackelberg
+leader** uses to manipulate forward-looking decisions of a **Stackelberg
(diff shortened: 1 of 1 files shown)


arnavs:
@ajozefiak We might not be able to get an external review on this. So if you could compare this, whenever you have time, to the one online on the lectures.quantecon.org/py, and check that our tests are solid and that there are no obvious differences, that would be great.

PR #561, opened 2019-06-05
June 2019 Fixes

arnavs:
Fixes based on the latest round of Dockerized execution.


arnavs:
Merging as it contains a bugfix. There are other tests whose values we can tweak. But nothing fatal.

PR #569, opened 2019-06-21
BUG: updating badges url

AakashGfude:
ipynb badges url from the local path, pdf badges url from the global path


mmcky:
Note: This means that `pdf` download will only work in the production environment on s3 and will not work when viewing locally.

PR #571, opened 2019-06-26
FEAT: Updating python pdf download in template file for consistency 

AakashGfude:
for julia, at present it is a disabled button


mmcky:
thanks @AakashGfude -- @arnavs @jlperla we will enable the `Julia` complete pdf download once we have compiled a `Julia` pdf file.

PR #577, opened 2019-07-09
FEAT: Make commands for better user experience

AakashGfude:
Make command structure in this PR :-

`make setup` - installs the dependencies

`make jupyter` - converts rst to ipynb (unexecuted)

`make website` - converts rst to website (along with execution and conversion to html)

`make website parallel=True` - parallel version of make website

`make preview` - opens the unexecuted notebook folder (_build/jupyter)

`make preview lecture=code_blocks` - opens the unexecuted lecture (code_blocks.ipynb in this case)

`make preview target=website` -creates a python server and serves the _build/jupyter_html folder

This PR will rework the default settings in the conf.py file to enable make jupyter to focus on building local notebooks. The settings required for build the website will be enabled using override flags from the cmd line built into the make website command.

We have also added separate build pathways. Local editing notebooks will remain in _build. The files used for constructing the website including ipynb, executed, html, _static are all now contained in _build_website. The files used for coverage ipynb, executed-ipynb are in _build_coverage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 _build/
 _build_website/
 _build_coverage/
+dask-worker-space/
 .DS_Store
 **/.ipynb_checkpoints/
 **/*.ipynb
--- a/Makefile
+++ b/Makefile
@@ -7,24 +7,22 @@ SPHINXBUILD   = python -msphinx
 SPHINXPROJ    = lecture-source-jl
 SOURCEDIR     = source/rst
 BUILDDIR      = _build
+BUILDWEBSITE  = _build_website
 CORES 		  = 4
 BUILDCOVERAGE = _build_coverage
 
-
 # Put it first so that "make" without argument is like "make help".
-setup:
-	cd source/rst && ln -s ../_static _static
-
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
 
+# Install requiremenets for building lectures.
 setup:
 	pip install -r requirements.txt
 
 preview:
-ifeq ($(target), website)
+ifneq (,$(filter $(target),website Website))
 	cd _build/jupyter_html/ && python -m http.server
 else
 ifdef lecture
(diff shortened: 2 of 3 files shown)


jstac:
Hi guys.

Regarding `make local`, I like your suggestion @AakashGfude.  In any case, `make local` should be dropped.

Second, I'm not sure that the structure of the `_build` directories is logical.  It would be confusing for a user to see `_build` and `_build_website` at the same level.  Let's avoid any structure where an intelligent new user would look at it and say "huh??".

I suggest instead one `_build` folder with directories inside such as `jupyter`, `website`, etc.

mmcky:
@jstac indeed - there is no `make local` on the Julia side as we never set one up so it won't be in this Makefile. @AakashGfude I'll amend your description above to remove confusion. 

The main reason we have the separate build folders for now is that the default `make clean` wipes everything in the `_build` directory. This is not ideal as the cache is wiped for every different workflow `jupyter`, `website, `coverage` etc. In this setup we can keep the different channels logically separate (cache preserved). It is also solely determined by settings in the `Makefile` as our extension will use the `BUILDDIR` settings that have been setup in the Makefile to write the folders, so none of these folder structures are hardcoded into the extension.

For those that will just be editing and writing lectures using `make jupyter` (and don't need to build the website) then they will just have a `_build` folder.

jstac:
@mmcky I understand --- but it's confusing for the user and illogical at the same time.  

I know I'm driving you guys crazy but I want the tools we build to be beautiful and logical.  So please adopt the convention I requested and adjust the behaviour of `make clean` to match.  

As far as I'm concerned, it's fine for `make clean` to wipe everything in `_build` and then for you to add options to further refine it's behaviour (according to what you guys need).

arnavs:
Sounds good, I will wait for your OK before reviewing/merging/etc. And thanks again for all the hard work on this toolchain. 

Only request from my side is to keep the docs up to date on common use cases after all changes and tweaks.

I think these are: 

* Looking at Jupyter output (currently `make jupyter` and `make jupyter-tests`.) For unexecuted output, I just hit `Ctrl+C` when Dask fires up, but maybe there is a better way. 

* Looking at HTML output. 

* Looking at coverage statistics/reports.

mmcky:
@arnavs the `Ctrl+C` will no longer be needed as `juypyter` will only focus on producing notebooks and not executing them. `make website` will build the full pipeline `ipynb` -> `exectution` -> `html`. I am working on updated docs now: https://github.com/QuantEcon/sphinxcontrib-jupyter/pull/200. 

I would like to go ahead and merge this as I have merged these changes in the `quantecon.build.lectures` after last nights testing over here. You will need to update the Docker to use `make website` or `make website parallel=True` and `make coverage` or `make coverage parallel=True`.

arnavs:
Sounds good to me. Merging now.

PR #584, opened 2019-07-19
FIX: fix issue with image urlpath prefix

### LINKED ISSUE #583: Incorrect Settings in Makefile
When migrating / reworking `Makefile` and `conf` in the `UX` changes -- an incorrect link was used in the `jl` Makefile for `jupyter_images_urlprefix`. It is currently set to reference the `py` site.

@AakashGfude let's review this morning and think if there are any other `py` specific items that need to be reviewed in the `Makefile` updates.

mmcky:
fixes #583

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,10 +47,10 @@ endif
 
 website:
 ifneq (,$(filter $(parallel),true True))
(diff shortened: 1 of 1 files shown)


mmcky:
@AakashGfude I am going to merge this PR to get the build going but if you could please double check the configuration that would be great. Thanks!

Issue #600, opened 2019-08-17
Check feasibility of using built in conda for notebook installation 


jlperla:
The key issue here is the difficulty in launching jupyter from the terminals... Requires adding the conda to path?

- [ ] check if easy way to get path to jupyter? Explain setup instructions to add to path.
- [ ] test installation process and steps
- [x] check if jupyterlab included by default
- [x] easy way to create bookmark to cloned Julia stuff?

jlperla:
Seems to be a bug right now in IJulia with `jupyterlab(detached=true)` where it always asks to install via Conda?  `notebook(...)` does not do this  See #611

jlperla:
The response to the "bookmark to cloned julia" is that jupyterlab remembers the location.  So calling `jupyterlab(detached = true)` by default has decent behavior.

Jupyterlab is not installed by default.

PR #612, opened 2019-08-22
Instructions for using Julia with the built-in conda

jlperla:
Addressed #600 

@arnavs  Can you and someone else go through this, edit it as you see fit, verify links are correct, and test on our 3 target OSs?  I


arnavs:
Sure. Looking through this now. @XiaojunGuan- would be great if you take a look at it after I do my first pass. Will let you know when I'm done with that.

PR #642, opened 2019-08-29
Adjust config for theme and template path

mmcky:
This PR adjusts the configuration of `conf.py` and `Makefile` to correct the theme and template path. 

The `conf.py` file for the project was moved to the root level of the repository (which is the same as `lecture-source-py`) as we think that is a more logical location for it. The `source/rst` should now just contain the `rst` files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -c "./"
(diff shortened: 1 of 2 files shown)


mmcky:
@arnavs do you want to check if this branch fixes your issues with the theme?  You can run a quick build using `jupinx` 

```
jupinx -cws --files source/rst/getting_started_julia/getting_started.rst
```

PR #643, opened 2019-08-29
Compiled Binder

arnavs:
Changes for the precompiled binder environment. 

Tested with `jupyter-repo2docker .` on the `quantecon-notebooks-julia` that's spit out.


arnavs:
Closing this since we will be rethinking based on the new PackageCompiler approach (`sd-notomls`.)

PR #644, opened 2019-08-29
fix parallel interface with jupinx and allow integer specification

mmcky:
This PR allows you to specify the number of workers rather than `True/False` to trigger parallel builds.

```
jupinx -cws --parallel=8
```

or 

```
make website parallel=8
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ SPHINXPROJ    = lecture-source-jl
 SOURCEDIR     = source/rst
 BUILDDIR      = _build
 BUILDWEBSITE  = _build/website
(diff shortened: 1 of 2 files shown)


arnavs:
Is there still an upper bound from `zmq`?

mmcky:
hey @arnavs yes -- https://github.com/QuantEcon/sphinxcontrib-jupyter/issues/261 is still an issue. So the build server runs sequential for robustness.

arnavs:
Roger that. Looks good to me. But I'll give it a go for good measure before merging.

Issue #654, opened 2019-09-01
Launch Button Default to Binder


arnavs:
cc @DrDrij @jlperla @mmcky 

We changed the order of the options in the launch bar (which took.) The issue is that the actual bar still defaults to `CoLab` (which doesn't exist anymore, as an option.)

Any advice? @DrDrij could you take a quick look through our `julia.tpl` and see if there's an easy fix?

<img width="926" alt="Screen Shot 2019-09-02 at 12 05 51 AM" src="https://user-images.githubusercontent.com/5009477/64082824-950a5380-cd15-11e9-9859-41ea6a8bc6d8.png">

jlperla:
To make sure I understand: is it that `colab` in https://github.com/QuantEcon/lecture-source-jl/blob/master/theme/templates/julia.tpl#L167 should be replaces by the string from https://github.com/QuantEcon/lecture-source-jl/blob/master/theme/templates/julia.tpl#L205

jlperla:
@DrDrij  If you can increase the delay on the popup of the launch button by a bit more when you look at this, I think it may be slightly better, but I defer to you on usability.

DrDrij:
@jlperla You correctly identified the lines to amend. I've also increased popup opening delay from 200ms to 500ms. The PR is ready here:
https://github.com/QuantEcon/lecture-source-jl/pull/658

arnavs:
Fixed in the latest clean build.

PR #658, opened 2019-09-02
NB Launcher - replace colab with myBinder, increase popup delay #654

DrDrij:
- Removed and replaced colab launcher server option with myBinder
- Increased time for Notebook Launcher opening on hover from 200ms to 500ms


arnavs:
@jlperla Sounds good to me, merging and then we can confirm with the next test build.

PR #659, opened 2019-09-02
Remove Compat [WIP]

arnavs:


arnavs:
There is no coverage regression. Merging now. 

cc @jlperla

Issue #673, opened 2019-09-04
Change kernelspec displayname to include minor version, and check nb


jlperla:
There are unnecessary conflicts introduced for people executing a notebook.   The git diffs, and deployment of updates from the server, will be easier if we can minimize these changes.

See 
![image](https://user-images.githubusercontent.com/13106721/64258743-9e91e800-cedc-11e9-9728-de154a56cac2.png) 

Which you can see only has irrelevant changes.  In particular:
- Can we change it to ` "display_name": "Julia 1.2.0",`.  As I said before, I think that the Julia lecture breadcrumbs should not have minor versions at all if possible (i.e. Just say Julia, not Julia 1.2), so this shouldn't effect the theme.
- Is the actual nbformat_minor 2, or is `"nbformat_minor": 4` more appropriate given the released version we are using?  If so, lets bump it in the generated jupinx.

@mmcky I assume this is a jupinx thing.

mmcky:
@DrDrij can we make this change to the breadcrumbs for `julia`?

jlperla:
@mmcky Just to be clear, the real pain right now is that anytime you execute a file, it changes those lines in the output (which makes it have false-positives on git changes).  This makes it hard to update notebooks, even when nothing changes.

If the kernelspec needs to display Julia 1.2.0 or whatever in the theme for now, then so be it.  But I think it would be better in the medium-run if it just displayed "Julia" like the python lectures just display "Python" rather than a major/minor version.  Julia's semver is pretty solid these days (even if the packages are in chaos!) so the minor version isn't very important.

arnavs:
@DrDrij  said he can look at this for a bit now, and then in a few hours. ETA is 1/2 day for patching the theme.

jlperla:
Not worth patching the theme.  But is there any way to change the `display_name` and `nb_format_minor` so it has less git issues?

arnavs:
Sorry @jlperla, by theme I meant `julia.tpl`, i.e. the template file we're using to generate the HTML. Not the `jupyterlab` theme. 

Closing because of @DrDrij's PR in #678

jlperla:
Before this is closed, I just want to verify that this is fixed for the generated jupyter notebooks, which was the issue I raised. Ie that the kernelspec numbers in the config are set properly. 

Can we check a few os to make sure the jupyter notebook doesn't trigger a git diff after executing a cell that doesn't change any output? I will do on windows jupyter and jupyterlab

[...]

PR #678, opened 2019-09-06
Hard code breadcrumbs, remove reference to kernel spec #673

DrDrij:


arnavs:
Looks good to me. #673

Issue #634, opened 2019-08-28
Wait on RegressionTables and Re-Add to Lectures


arnavs:
This is currently commented out, and removed from Project/Manifest. Before 1.1, we should take another look.

jlperla:
For https://github.com/jmboehm/RegressionTables.jl/issues/26

Is there any way we could make the PR ourselves?

arnavs:
Moving this to 1.1 since we decided to wait on Johannes.

jlperla:
With this fixed, lets test and see if we can get this and the dataframes stuff bumped up, and turn in back on in the lecture notes. I would love to teach with FixedEffectsModels and this in a few weeks.

arnavs:
Sounds good. Will do as soon as Johannes tags a release. Which should be soon, since I think the FixedEffects upgrade is fairly minor.

arnavs:
@jlperla I'm closing this because #681 looked good.

PR #681, opened 2019-09-09
Reinstate RegressionTables, FixedEffects into Lectures

arnavs:


arnavs:
Yeah, this is looking good. 

Close #634.

PR #689, opened 2019-09-13
Essentials and Fundamental Types updates

jlperla:


jlperla:
@arnavs  Yours to review and doublecheck

As this adds in some new exercises, best to check the output to make sure the layout is good.

PR #692, opened 2019-09-17
WIP: In progress on about julia lectures

jlperla:
@mmcky  I am making progress on this, but not there yet.  When do I need this done by? Do I have a few more days?

### DIFF
--- a/source/rst/about_julia_lectures.rst
+++ b/source/rst/about_julia_lectures.rst
@@ -1,0 +1,193 @@
+About these Lectures
+=====================
+
+Overview
+--------
+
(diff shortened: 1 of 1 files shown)


mmcky:
@jlperla and @arnavs would it be ok if we moved this to a new branch called `standalone-bucket` and then we can put all changes relating to the `ec2` migration into that branch. This will make testing and deployment much simpler than individual PRs. Would that be ok?

mmcky:
I setup this one on the python side: https://github.com/QuantEcon/lecture-source-py/pull/683

jlperla:
Sure, modify or move to whatever extent you wish

PR #694, opened 2019-09-20
minor correction in math environment in perm_income.rst

AnjuJoon:


AnjuJoon:
Hi @arnavs , latex gives an error when it sees `begin{align}`, `begin{gather*}`, `eqalign`, etc. in math environment and stops. It is more comfortable with`begin{aligned}` to align multiline equations. It is probably due to the packages used in the preamble.

arnavs:
@AnjuJoon I see, thanks for explaining. Makes sense that we didn't encounter this error, since we have generally not been working with the PDFs.

PR #695, opened 2019-09-20
minor change in math environment in wald_fridman.rst

AnjuJoon:


arnavs:
Will merge this one, since nothing breaks and there's no reason to hold up the whole queue.

PR #703, opened 2019-09-24
Add ApproxFun, BlockBandedMatrices, LightGraphs

arnavs:
Running coverage on this now. 

If it looks good, we could also use this branch for the content involving these packages


arnavs:
Coverage looked good last I ran it, so I think this is ready for preview. And then we can develop content based on LightGraphs etc. (see also #228 that @XiaojunGuan is working on.)

cc @jlperla

PR #704, opened 2019-09-25
fullstops added to 5 lectures Batch(1)

AnjuJoon:


source/rst/dynamic_programming/career.rst
```diff
@@ -322,14 +322,14 @@ Exercise 1
 
 Using the default parameterization in the ``CareerWorkerProblem``,
 generate and plot typical sample paths for :math:`\theta` and :math:`\epsilon`
-when the worker follows the optimal policy
+when the worker follows the optimal policy.
 
 In particular, modulo randomness, reproduce the following figure (where the horizontal axis represents time)
```
arnavs: @AnjuJoon Do we know why this wasn't picked up?

jlperla:
@arnavs we need to be super careful with the review and merge since this was a script. Also instead of pointing out problems with the script, it means we just need to fix them.

arnavs:
@jlperla For sure. Since I don't have access to the script, I figured I would just highlight the issue for now.

source/rst/dynamic_programming/career.rst
AnjuJoon: hi @arnavs this line is not picked up because it refers to some figure below it for example in this line "the following figure ..."

The script is such that if the line refers to some code below or figure or math equation just next to it then it will not add a full stop after that line. 

If you want that to be changed then it's not a problem, I can do that. Thanks for merging

AnjuJoon:
Hi @jlperla and @arnavs thanks for merging. I would make sure to be very careful with these. Since I know the script so I also know the possible bugs and it would be easy for me to debug. Thanks for merging. I would be making more PRs at the weekend. Thanks in advance for bearing the trouble to review those.

arnavs:
Sounds good. And thanks @AnjuJoon for your hard work, as always.

For this one, I manually skimmed before merging. But going forward I will go over it with a finer comb.

AnjuJoon:
> Sounds good. And thanks @AnjuJoon for your hard work, as always.
> 
> For this one, I manually skimmed before merging. But going forward I will go over it with a finer comb.

Thanks @arnavs 1 ` :- )`

PR #712, opened 2019-09-26
Various Fixes

arnavs:
WIP for now, will change the labels when I think this is complete.


arnavs:
Merging this as there is some stuff we need to see in the preview.

PR #714, opened 2019-09-28
contents:: :depth:2 added at right place

AnjuJoon:


source/rst/multi_agent_models/markov_perf.rst
```diff
@@ -11,6 +11,8 @@ Markov Perfect Equilibrium
 .. index::
     single: Markov Perfect Equilibrium
```
AnjuJoon: @arnavs @jlperla this `.. index::  ` is almost under every title in this lecture but it is not so in other lectures. Is this deliberate?

arnavs: @AnjuJoon Don't think so; looking at the blame for this lecture, these were in there since the initial commit last year.

PR #721, opened 2019-09-28
Fix Failing Regression Test

arnavs:


arnavs:
There is a floating-point comparison in lake model that passes (on the server) at tolerance `1e-6` and fails at `1e-7`. This PR contains the fix so it doesn't falsely show up as a failing lecture.

PR #723, opened 2019-09-30
Added assignment for Generic Programming

jlperla:
@arnavs I think this is ready to be checked and merged.


source/rst/more_julia/generic_programming.rst
```diff
@@ -738,6 +738,65 @@ The main way to implement this in a generic language is with a design approach c
...
+    using LinearAlgebra
+    function trap_weights(x)
+        return step(x) * [0.5; ones(length(x) - 2); 0.5]
+    end
+    x = range(0.0, 1.0, length = 100)
+    ω = trap_weights(x)
+    f(x) = x^2
+    dot(f.(x), ω)
+    
+However, in this case the creation of the ``ω` temporary is inefficient as there are no reasons to allocate an entire vector just
```
arnavs:
```suggestion
However, in this case the creation of the ``ω`` temporary is inefficient as there are no reasons to allocate an entire vector just
```

PR #726, opened 2019-09-30
remove slash from labels in rational_expectations.rst

AnjuJoon:
@arnavs @jlperla Ipynb translator does not acknowledge this special character in labels


AnjuJoon:
Thanks @arnavs

PR #693, opened 2019-09-18
UPDATE: adjust structure for new julia.quantecon.org s3 bucket 

mmcky:
This PR makes adjustments for the new `julia.quantecon.org` s3 bucket structure. 

**Tasks:**

- [x] add new landing page
- [x] add `about` page
- [x] update `julia` theme to new independent theme adjusted for new hosting context

PR #727, opened 2019-09-30
Translated geom_series.py to Julia 1.2

yolsever:


arnavs:
Thanks @yolsever for turning this around so quickly and with a good amount of detail. I'll have a look at this today, but we will probably consider new content like this after we finish the URL migration in #693

arnavs:
Replacing this PR with one in a local branch that we can edit.

PR #744, opened 2019-10-04
add link to spline explain interpolation

deivisangeli:
most students probably are not familiar with splines


arnavs:
Confirmed that this builds right, merging now.

PR #752, opened 2019-10-09
Various Fixes (2)

arnavs:


arnavs:
Running coverage on this now. This is what we had before. Also want to see this on the server, so merging.

PR #757, opened 2019-10-14
FIX: update julia_index to quantecon-jl for new org site

mmcky:
This PR updates some broken links due to migration of org site


arnavs:
Thanks @mmcky

PR #760, opened 2019-10-15
math correction for pdf generation

AnjuJoon:


mmcky:
@arnavs if you're happy with this PR I will re-run the `pdf` for `julia` once it is merged.

PR #771, opened 2019-10-23
Replace theme with minimal, CSS, JS and template files

DrDrij:
This branch replaces the QE Julia Lectures theme and associated files with a generic "minimal" theme.
- Replaced julia-html.tpl with html.tpl
- Replaced css and js files with a base.css and base.js
- Removed redundant image assets


arnavs:
Probably you're working on this, but the launch bar order etc. is currently hard-coded into the `julia-html.tpl` file. I don't see that stuff in the new `html.tpl`; where would we put it?

mmcky:
just to clarify. The idea of this `minimal` theme is designed to support local editing and alllow focus on the content. The idea being we could take the qe theme private which could be build `locally` with some minimal setup or via the `build` instance.

jlperla:
@mmcky  I am just doing some PR cleanups now and managing tasks.  Is this still relevant?  For the most part it seems like there isn't that much benefit in these sorts of tasks prior to the jupyterbook migration (whenever that may be!)

mmcky:
@jlperla so on the `python` side we have added a `minimal` theme to our public repositories and made the `quantecon` theme `private`. This PR adds the `minimal` theme. I should review this to make sure it is up to date before merging. 

@DrDrij would you mind casting an eye over this to see if it is up to date?

We have added a theme repo for each lecture set so the julia theme repo would be `lecture-python-jl.theme`

jlperla:
@mmcky  For sure.  We will follow your lead on this.  Now is a good time to go through these issues.  If it made it easier for me to do local edits, then so much the better!

For the theme, would the jupyter launch bar stuff be here or somewhere else?  Does this PR change where to add colab support/etc?   Obviosuly, that stuff doesn't make any sense for local edits.

mmcky:
Here are some previews of the `minimal` theme

image

TOC

image

Lecture

image

jlperla:
@mmcky  I love it.  So this is what it would look like with local builds now?

mmcky:
Once the PR is merged - yes. Just need to coordinate

- [ ] setup `lecture-source-jl.theme` and populated with current `theme` files
- [ ] adjust build instance to build using private theme by updating [build scripts](https://github.com/mmcky/QuantEcon.aws.admin)

mmcky:
@DrDrij given the age of this PR could you please review and give me your tick of approval. In the meantime I will do the adjustments listed above.

mmcky:
- [ ] [This PR](https://github.com/mmcky/QuantEcon.aws.admin/pull/84) needs to be merged at the same time. @arnavs can you review?

mmcky:
@DrDrij there seems to be some missing asset elements (i.e. icons). See preview of front page here: 

image

arnavs:
@mmcky Looks good to me. Added the THEMEPATH PR to admin and I think this one is good to go. 

Geronimo.

mmcky:
This is missing some elements in the `generic` theme and was waiting for @DrDrij to review. But given its never published this should be ok.

PR #772, opened 2019-10-26
Translated geom_series.py to Julia 1.2

arnavs:
From @yolsever's #727.


arnavs:
@yolsever I think this is almost complete. Comments: 

1. Let's cut down on the broadcasting (i.e., use of `.+`, `./`, `.-`, `.*`) to cases where it's necessary. Otherwise it makes the code hard to follow (what is a vector, what is a scalar, etc.)

2. There's a line with a comment marked `### not sure about this line`. The line looks good and clear to me. Check and remove the comment if appropriate. 

3. Do a final [style guide](https://github.com/QuantEcon/lecture-source-jl/blob/master/style.md) pass. Nothing struck me as wrong when I skimmed, but good to check. I think the main points here would be: 

> 5. Maintain this correspondence between math and code even if the code is less efficient. Only optimize if it is really necessary.

> 6. Ensure that all code can be copied and pasted without modification into functions for performance and modification without changes to scoping (e.g. no local or global ever required)

Once this is done, we can do a formal review on this.

arnavs:
Note: If the equations are written in scalar form, but they vectorized the `python` side, it may be clearer for us to use list comprehensions or something. e.g.: 

```
discounted_values = [b^t * p(t) for t in 1:T]
present_value = sum(discounted_values)
```

or something like that. 

But if you do go this route, make sure you follow the style guide points about comprehensions/iterations.

jlperla:
I believe, if you are only doing sums, that you can even try things like
```julia
present_value = sum(b^t * p(t) for t in 1:T)
```
If I recall... Comprehensions can be very clear. 

But we definetely need to remove every single broadcast that isn't necessary

PR #779, opened 2019-11-06
Decrease math size to match surrounding text

DrDrij:


DrDrij:
Needs further review.

PR #796, opened 2020-01-28
Julia 1.3 Final Updates

arnavs:


arnavs:
PR is good to go on all lectures. Just need to tweak setup instructions.

arnavs:
@jlperla Merging this now. Removing `Queryverse` struck `PyCall` from the manifest. So we can let the server at ANU do our final round of testing, and if things look good on the preview, then we are good! 

(After we update the notebooks etc., of course.)

PR #801, opened 2020-03-04
fix minor label error in Aiyagari

Shunsuke-Hori:
I happened to find a minor label error in Aiyagari lecture. 

This PR fixes it.

image


arnavs:
Thanks @Shunsuke-Hori

PR #805, opened 2020-04-09
Better spacing for web layout

jlperla:
@arnavs Can you check this out and merge if you think it works? Maybe look for any other places that it might show up


arnavs:
Sure. Will (1) merge this, (2) do a quick scan-through for other places, (3) trigger a preview build.

jlperla:
Maybe run the 2 new lectures as well to make sure there are no major package issues there. Still dream of getting those ready in a month or two

PR #837, opened 2020-05-15
An tweak based on my experiences to colab instructions.

jlperla:
@arnavs See my attempt at this.  Merge, edit, and test as you see fit.

I think telling people to always refresh the browser is the most predictable even if it is perhaps unnecessary?


arnavs:
Yeah, refreshing works. Might as well do that, since it's more idiot-proof and familiar.

PR #853, opened 2020-05-19
Fix parameters of function T(...) (Bellman operator)

artemsolod:
1) global `grid_y` changed to parameter `grid`;
2) unused parameter `Tw` removed from definition and subsequent calls.

Please not, that I did not make local build of the lecture, however, I did run the code within the appropriate environment.


jlperla:
Thanks so much.  We will take a look but it looks like good to me.

arnavs:
Yeah, this checks out for me too. Will merge and pass it to coverage bot.

PR #857, opened 2020-05-22
FEAT: Added handling for image/png files

AakashGfude:
This should handle image/png format in pdf. we can similarly handle other image formats.

@mmcky I had also specifically added the `out` block , which is not in python files. I can remove it for consistency, or let it be here ?


mmcky:
I think it should be contained in `Out` blocks. 👍

mmcky:
I will cc this to `lecture-source-jl.theme` as well.

PR #861, opened 2020-06-02
Copy edits for Numerical Linear Algebra

jlperla:


jlperla:
@arnavs  Review carefully, merge, and then go through the generated HTML/etc. to make sure there were no regressions in the display/etc.

Issue #859, opened 2020-05-22
Thoughts on covid and SciML lecture outline


jlperla:
@jstac @thomassargent30  @ChrisRackauckas 

Starting with the python version as a reference: https://python.quantecon.org/sir_model.html

My thought is that there could be two lectures here:
- `Modeling Covid 19/Introduction to SDEs`
- `Policy Uncertainty with Covid 19/Introduction to Scientific Machine Learning`

Here are some rough thoughts, none of which I am especially attached to.  But I think a good goal is that everything in the "first lecture" are things where the code is probably already complete somewhere so it could be done quickly.

For the first lecture:
- We could start by porting a good portion of https://python.quantecon.org/sir_model.html to the OrdinaryDiffEq package,  up to the Ending the Lockdown.
- We then could add a section introducing the basic structure of a SDE that are initial value problems (i.e. forward equations, no control problems in these lectures)
- As an application of that, we add in a reasonable additive aggregate shock to the SIR model.  There are a few candidates of the top of my head:
  - We could make the `i` have an additive shock representing people migrating into the region?
  - We turn it into an ODE with 4 states... S, I, R, sigma.  Where the sigma was the infection rate.  Then we make the sigma stochastic.
  - Similarly, we could make the beta (i.e. the transmission rate) stochastic, since I think there are aggregate and unpredictable changes to how much people "follow the rules".
  - Chris, maybe you have some ideas from reasonable things people have done before and where there is code?
- Given the SDE, we show simulating a path of it, as well as ensembles using the diffeq tools.  We can focus on the typical quantiles (e.g. mean, 5/95% etc.).
  - For example, I think it would be very useful to look at the aggregate deaths to date at some point `T` in the future?  I think in the simple setup there, it is a constant times the `R`. That parameter could be useful for the second lecture since nobody really knows it.
- Finally, I think it is worthwhile for economists to learn the connection between the SDEs for the Langevin equations typically used in the approximation of chemical reactions/etc.  
  - For that, a discrete number of agents bumping into each other can be approximated by an SDE.
  - There, the noise in the SDE captures the discreteness, which could be useful for all sorts of models as well as modeling covid in small groups or isolated regions.

For the second lecture:
- There are a lot of opportunities here to introduce tricks from scientific machine learning as well as modeling uncertainty.
- We could take the "Ending the Lockdown" example from the python ones as a starting point.
[...]

PR #867, opened 2020-06-27
WIP: Covid sde

jlperla:
This is a work in progress, but it is time to start examining it.  There has been virtually no copy editing/etc.

@jstac  @ChrisRackauckas  @thomassargent30


jstac:
Implements some of the discussion in #859

jstac:
Very interesting.  I added only minor corrections.  The stochastic time series map better to the data than the deterministic ones.  So there must be significant aggregate shocks,

covid


I'll review more carefully when I have time but everything is clear and nicely explained.

ChrisRackauckas:
I plan to go through this in detail, but let's say this weekend. I had a paper resubmission come up.

jlperla:
@jstac  The only real chance on this PR from the one you modified is that I move this into its own section on continuous-time - where we can build up the other lecture slowly, and I moved the raw notes at the bottom intended for the 2nd lecture into #868

source/rst/continuous_time/seir_model_sde.rst
```diff
@@ -0,0 +1,913 @@
...
+
+.. code-block:: julia
+
+    p_gen = @with_kw ( T = 550.0, γ = 1.0 / 18, σ = 1 / 5.2, η = 1.0 / 20,
+                      R̄ = 1.6, B = (t, p) -> p.R̄, m_0 = 0.01, N = 3.3E8)
+
+Note that the default :math:`B(t)` function always equals :math:`\bar{R}`
+
+
+Setting initial conditions, we will assume a fixed :math:`i, e`, :math:`r=0`, :math:`R(0) = \bar{B}`, and :math:`v(0) = 0.01`
```
ChrisRackauckas, line 301: Not sure what `v` is supposed to be referring to.

jlperla: Sorry, not on computer, but it might be old notation for the death probability in the python ? Maybe I changed it to `m(t) ` for mortality?

source/rst/continuous_time/seir_model_sde.rst
```diff
@@ -0,0 +1,913 @@
...
+------------------------------
+
+Consider that the effective transmission rate :math:`R(t)` could depend on degrees of randomness in behavior and implementation.  For example,
+
+* Misinformation on Facebook spreading non-uniformly
+* Large political rallies, elections, or protests
+* Deviations in the implementation and timing of lockdown policy within demographics, locations, or businesses within the system.
+* Aggregate shocks in opening/closing industries
+
+To implement this, we will add on a diffusion term to :eq:`Rode` with an instantaneous volatility of :math:`\zeta \sqrt{R}`.  The scaling by the :math:`\sqrt{R}` ensures that the process can never go negative since the variance converges to zero as R goes to zero.
```
ChrisRackauckas, line 545: Not sure it actually ends up as a provably positive process. At least the argument and intuition aren't necessarily correct. `u' = -sqrt(abs(u))` is an example where the rate goes to zero but this process hits exactly zero in finite time, at which point its solution is non-unique but it's either 0 or it goes negative. I think the fact that the diffusion length scales by another sqrt might make this work out, but it's not exactly trivial.

source/rst/continuous_time/seir_model_sde.rst
```diff
@@ -0,0 +1,913 @@
...
+    \begin{aligned}
+    diag(G) &:= \begin{bmatrix} 0 & 0 & 0 & 0 & \zeta \sqrt{R} & 0 & 0 & 0 \end{bmatrix}
+    \end{aligned}
+
+
+Since these are additive shocks, we will not need to modify the :math:`F` from our equation.
+
+First create a new settings generator, and then define a `SDEProblem <https://docs.sciml.ai/stable/tutorials/sde_example/#Example-2:-Systems-of-SDEs-with-Diagonal-Noise-1>`__  with Diagonal Noise. 
+
+We solve the problem with the `SRA <https://docs.sciml.ai/stable/solvers/sde_solve/#Full-List-of-Methods-1>`__ algorithm (Adaptive strong order 1.5 methods for additive Ito and Stratonovich SDEs)
```
ChrisRackauckas, line 580: this is not an additive noise SDE

source/rst/continuous_time/seir_model_sde.rst
jlperla: I added in the scaling by the sqrt when we discussed this previously, and you had suggested it was a good way to keep things positive.

But if I have learned anythign over the years, it is that continuous time stochastic processes are is often counterintutive, and that the details matter.

For this one, I think the finance crew calls it the CIR process: https://en.wikipedia.org/wiki/Cox%E2%80%93Ingersoll%E2%80%93Ross_model

It says there that the scaling avoids the possibility of negative interest rates for all `a` and `b`, and an absorbing state of zero is precluded as long as the drift is strong enough relative to the variance.  I think this comes out of the ergodic distribution being Gamma for the right combination of parameters: https://en.wikipedia.org/wiki/Cox%E2%80%93Ingersoll%E2%80%93Ross_model#Distribution

With that said, I don't think it means that the SDE simulators would necessarily always keep things positive?  Which might be what I ran into later in the code (which you might see in https://github.com/QuantEcon/lecture-source-jl/blob/459e806e3acaa3371013112ed10eae7c5f38ce4a/source/rst/continuous_time/seir_model_sde.rst#implementing-the-complete-system  where I fudge the variance to avoid a sqrt of a negative.  Probably better ways to do this.

I think a little adhoc on the SDE side for this is OK as long as we are clea rwhat is going on.

source/rst/continuous_time/seir_model_sde.rst
jlperla: I see.  This is because of the scaling of the variance?  If so, maybe putting in a note on what the general goto algorithms are for additive vs. other noise couldn't hurt.  Economists (and definetely me!) are utterly ignorant of this stuff.  Everything is homegrown Euler-Maruyama at best.

ChrisRackauckas: Yes, additive noise is what dW is only multiplied by constants. Here it's just diagonal noise.

source/rst/continuous_time/seir_model_sde.rst
ChrisRackauckas: >It says there that the scaling avoids the possibility of negative interest rates for all a and b, and an absorbing state of zero is precluded as long as the drift is strong enough relative to the variance.

Alright, cool. I think we should mention that the explanation given is really just a heuristic, and to know it's going to be positive requires more details. 

>With that said, I don't think it means that the SDE simulators would necessarily always keep things positive?

Yeah, being that close to zero can cause numerical difficulties

jlperla: Comment added

source/rst/continuous_time/seir_model_sde.rst
```diff
@@ -0,0 +1,914 @@
...
+                 σ*e - γ*i;     # di/dt =        σe -γi
+                       γ*i;     # dr/dt =            γi
+                ]
+    end
+
+Written this way, we see that the four equations represent the one-directional transition from the susceptible to removed state, where the negative terms are outflows, and the positive ones inflows.  The equation associated with the state :math:`R` has inflows and no outflows, so any agents which end up in that state are "trapped".
+
+Consequently, with no flow leaving the :math:`dr/dt` and strictly positive parameter values the ``R`` state is absorbing, and :math:`\lim_{t\to \infty} r(t) = 1`.  Crucial to this result is that individuals are perfectly divisible, and any arbitrarily small :math:`i > 0` leads to a strictly positive flow into the exposed state.
+
+We will discuss this topic (and related topics of irreducibility) in the lecture on continuous-time Markov chains, as well as the limitations of these approximations when the discreteness becomes essential (e.g. continuum approximations are incapable of modeling extinguishing of an outbreak).
```
jstac, line 159: Perhaps this is not true once aggregate shocks are added, if, say, the paths admit jumps.  Might be worth qualifying?

source/rst/continuous_time/seir_model_sde.rst
```diff
@@ -0,0 +1,922 @@
...
+transitions from :math:`R \to S`.
+
+The transitions between those states are governed by the following rates
+
+* :math:`\beta(t)` is called the *transmission rate* (the rate at which individuals bump into others and expose them to the virus).
+* :math:`\sigma` is called the *infection rate* (the rate at which those who are exposed become infected)
+* :math:`\gamma` is called the *recovery rate* (the rate at which infected people recover or die).
+
+
+In addition, the transmission rate can be re-parameterized as: :math:`R(t) := \beta(t) / \gamma` where :math:`R(t)` has the interpretation as the *effective reproduction number* at time :math:`t`.  For this reason, we will work with the :math:`R(t)` reparameterization.
```
jlperla, line 112: @jstac  I took this from the python lecture, but I am not sure it is correct in epidemilogy lingo.  This might be the R_0 instead (i.e. the "Basic reproduction number") instead.  See https://en.wikipedia.org/wiki/Basic_reproduction_number and other stuff.

But understanding from this is that the base reproduction number is that where we make everyone susceptible, while the effective reproduction number takes into account the current equilibrium?

jlperla: I believe in your implementation you may talk about `R0` instead? https://python.quantecon.org/sir_model.html#Implementation

jlperla: Finally, for the "effective reproduction rate", which might actually be useful to plot as well, I think it might be simply `R_e(t) = R_0 * s(t)`  i.e. the derivative of the  `R_0 * I/N * S` flow wrt `I`.  Or the marginal number of new infections for an increase in the number of infected?

source/rst/continuous_time/seir_model_sde.rst
jlperla: I spent a bunch of time looking into this and clearing this up.  This turns out to be a very subtle point, but what I had written before was correct.  It would only be true if R_0 > 0 for this sort of model, I think.

The latest updates though have changed a lot so this won't be an issue.

source/rst/continuous_time/seir_model_sde.rst
jlperla: @jstac   I implemented these sorts of changes in the latest version, and there are a significant number of changes along these topics.

jlperla:
@jstac  @ChrisRackauckas @thomassargent30  

I just pushed a major set of changes 

The summary of changes is:
- I moved the jump diffusion stuff to a do later.  There was enough content for the SDE stuff that it would be overwhelming otherwise. 
- Also, I split the deterministic and SDE versions into separate lectures for the same reason.  It will also mean we don't have to fully nest the determinstic and SDE models.
-  One benefit of this is that without the jump-diffusions, it will be much easier to push these lectures out since they don't require deploying new packages.
- The SDE model now uses an SIRD model.  This lets us teach people about a different type of covid approach (and it is referenced in a paper by jesus and chad), and simplifies those equations a bit.
- There is a lot of new material on the `R_0` and it is now "correct" (or at least consistent with much of the literature).
- There are a bunch of new experiments and results in teh SDE lecture.  At this point, I think the lectures are complete.

I am going to merge this PR because it is the only way to initiate a preview build for quantecon.  We can then review the preview and do any copy editing.  Or we do it in waves (i.e. get it live, then copy edit over time).

PR #870, opened 2020-07-12
COVID SDE lecture edits

ChrisRackauckas:


ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/issues/548 is what was missing from the plots with output_func. Just to get things done, I avoided that for now.

PR #881, opened 2021-04-23
It is a very bad habit to use non const globals

azev77:


jlperla:
There are all sorts of places where closures are used.  This is not intended to be for performance but rather for interactice scripting style.  Inside of a function these things are fine, and I think I tell people to copy code inside a function anytime you want to do something for performance.

azev77:
The first piece of advice in Julia's Performance Tips: 
[Avoid global variables](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-global-variables)
-"Variables should be local, or passed as arguments to functions, whenever possible."
Which is why I suggested passing `r` as an argument of the function `f`

jlperla:
This is code for exploratory scripting, not for performance.  If you put that code inside of a function (which you should always do) then it works fine.

That is, you should never, ever have code outside of a function for performance.  But this is a script.  If you avoid globals in scripts the entire codebase can be much more complicated and it isn't even encessary (since you would just copy paste stuff into a function and have performant closures.  That is: 
```julia
function what_i_would_run()
     r = 2.0
     f(x) = r * x * (1 - x)
    sol = fixedpointmap(f, iv=0.8)
end
```
is what you would do in reality.  That is the important point to make to people.  You avoid globals my always having things in functions, not by avoiding closures entirely.

